PR #24, opened 2015-10-11
CFY-3437 Simplified API

dankilman:


content/guide/agents-cloudify-agent.md
```diff
@@ -0,0 +1,188 @@
+---
+
+title: Agent Configuration
```
idanmo: Perhaps this page should be called: "Cloudify Agent" and a paragraph explaining what it is used for and then its configuration?
I just think that if you put all of this information in one place it is easier to understand..

content/guide/agents-cloudify-agent.md
```diff
@@ -0,0 +1,188 @@
...
+    type: cloudify.nodes.Compute
+    property:
+      agent_config:
+        # one of none, remote, init_script, provided
+        install_method: remote
+{{< /gsHighlight >}}
+
+Following are all the supported installation methods:
+
+* `none` - No agent will be installed on the host.
```
idanmo: Do you think it would be better to describe the prerequisites just under each mode listed here?

content/guide/agents-cloudify-agent.md
```diff
@@ -0,0 +1,188 @@
...
+## Pre-requisites for Init Script and Provided Agent Installations
+
+To use the `init_script` and `provided` installation methods, an IaaS plugin that supports it should be used. At the moment, only the openstack plugin supports
+these installation methods.
+
+
+# Configuration Locations
+
+Agent configuration is composed of several locations which all adhere to the same schema. The schema is based on the `cloudify.datatypes.AgentConfig`
+which is defined in the standard `types.yaml`.
```
idanmo: on the ... `datatype` which...

content/guide/agents-cloudify-agent.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Pre-requisites for Init Script and Provided Agent Installations
+
+To use the `init_script` and `provided` installation methods, an IaaS plugin that supports it should be used. At the moment, only the openstack plugin supports
+these installation methods.
+
+
+# Configuration Locations
+
+Agent configuration is composed of several locations which all adhere to the same schema. The schema is based on the `cloudify.datatypes.AgentConfig`
```
idanmo: I think it is important enough to find a way to either put the AgentConfig data type here or somehow link to it.
What do you think?

content/guide/agents-cloudify-agent.md
dankilman: and remove the current general "top level" Cloudify Agent page?

idanmo: Yes, I think I would expect to find all the information I need about Cloudify agents in one page.
What do you think?

idanmo:
Its also needed to document this:
process_management:
  start_on_boot: true
for the init.d implementation (the default value is currently false but we'll change it to true).

dankilman:
It is not undocumented by accident. `start_on_boot` was broken last time I tested it. We should document this once it actually works

PR #64, opened 2015-12-11
CFY-4352 Initial snapshots overview plus some details and the install…

simon-bar:
…_new_agents workflow description


idanmo:
Should also include the process for creating a snapshot on 3.2.1 managers.

PR #70, opened 2015-12-14
added pip and virtualenv installation instructions

idanmo:


content/intro/installation.md
```diff
@@ -27,10 +27,12 @@ After the installation is complete, try running `cfy -h` in your terminal.
 
 ## Linux
 
-To install via rpm packages, you must have Python2.7.x and pip 1.5+ installed and Python2.7.x must be executable as `python` from the path.
+To install via rpm packages, you must have Python2.7.x, pip 1.5+ and virtualenv installed and Python2.7.x must be executable as `python` from the path.
```
nir0s: if we find out that we need these docs, you need to add it to content/installation/from-packages.md as well.

PR #89, opened 2015-12-24
CFY-4351 fixed broken links

idanmo:


content/apis/rest-service.html
```diff
@@ -9,6 +9,6 @@
 <html>
   <head>
     <meta http-equiv="content-type" content="text/html; charset=utf-8"/>
-    <meta http-equiv="refresh" content="0;url=http://docs.getcloudify.org/api/"/>
+    <meta http-equiv="refresh" content="0;url=http://cloudify-rest-client.readthedocs.org/en/3.3/"/>
```
nir0s: um?? that's the rest-service.html file. it should point to the rest-service API...

PR #102, opened 2016-01-11
CFY-4687 added openstack specific inputs and resources reference

nirbir:


content/manager/openstack-reference.md
```diff
@@ -0,0 +1,49 @@
+---
+layout: bt_wiki
+title: Openstack Bootstrap
```
nirbir: Probably needs a different name. 'Openstack inputs and resources'?

What about the file name?

nir0s:
Thank you. Will put this in once we decide on the location and naming (as you stated above) during the docs days.

PR #107, opened 2016-01-13
Add vsphere windows support details.

geokala:
Adds details of windows support currently being tested.


geokala:
No longer relevant.

PR #112, opened 2016-01-17
CFY-4723 edited plugin writing and creation pages

nirbir:


content/plugins/packaging-your-plugin.md
```diff
@@ -0,0 +1,26 @@
+---
+layout: bt_wiki
+title: packaging your plugin
```
nirbir: capital

PR #134, opened 2016-02-08
Cfy 4829 document implications of agent less mode

uric:


idanmo:
Please note that a 3.4-build branch was already created so you should probably cherry-pick.
Perhaps you should squash your commits first and then cherry-pick to 3.4.

PR #150, opened 2016-03-16
CFY-5050 - added prerequisites for OpenStack plugin

isaac-s:
And fixed a few typos

### DIFF
--- a/content/plugins/openstack.md
+++ b/content/plugins/openstack.md
@@ -17,9 +17,12 @@ For more information about OpenStack, please refer to: [https://www.openstack.or
 
 # Plugin Requirements
 
-* Python Versions:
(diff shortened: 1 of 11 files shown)


content/plugins/openstack.md
```diff
@@ -17,9 +17,12 @@ For more information about OpenStack, please refer to: [https://www.openstack.or
 
 # Plugin Requirements
 
-* Python Versions:
+* Python versions:
   * 2.7.x
-
+* Additional packages:
```
idanmo: Please use a note saying that if the plugin is installed from source, it requires these additional system dependencies.

isaac-s: Done.

PR #162, opened 2016-04-12
Add sudo prerequisite for agents

abramso:
Added a prerequisite for sudo


content/agents/overview.md
```diff
@@ -68,6 +68,7 @@ However, you should be aware of certain implications:
 
 * SSH port (22 by default) should be open for incoming connections.
 * An SSH server should be running on the host.
+* The agent user must have sudo privileges.
```
idanmo: sudo is required in general, not only for remote agent installation...

abramso: Not sure what's wrong with the wording, The "agent user" refers to the user running on the VM, not only for remote installation but also for running the agent services.

idanmo: The section "Pre-requisites for Init Script and Provided Agent Installations" should also contain the sudo requirement for installations on Linux.

abramso:
dd sudo prerequisite for Init-script

nir0s:
What's the status on this?

PR #181, opened 2016-05-28
CFY-5276 add native sudo support in fabric plugin

nir0s:


EarthmanT:
looks good. maybe also add the sudo_prefix input

nir0s:
I added it below..

content/plugins/fabric.md
```diff
@@ -31,14 +31,19 @@ As the fabric plugin is used for remote execution, the fact that it doesn't supp
...
 
 * Executing a list of `commands`.
 * Executing a Fabric task from a `tasks_file` included in the blueprint's directory.
 * Executing a Fabric task by specifying its path in the current python environment.
-* Executing a script by specifying the script's path or URL.
+* Executing a `script` by specifying the script's path or URL.
+
+## Executing commands or scripts with sudo
+
+Previously, to run `commands` or commands within a `script` using sudo, you had to explicitly write `sudo` where desired. Starting with fabric-plugin v1.4.1, The `commands` and `script` methods both accept a `use_sudo` input (which defaults to `false`). When true, the commands or script will be executed using sudo. This allows, for instance, to use the `sudo_prefix` fabric env property to run an alternative implementation of sudo. See below for examples on how to pass `use_sudo` and use the `sudo_prefix` property.
```
dankilman: I'm not sure this is the place to mention that this wasn't available previously.
Simply explain the feature

content/plugins/fabric.md
```diff
@@ -229,6 +237,7 @@ node_templates:
               host_string: 192.168.10.13
               user: some_username
               key_filename: /path/to/key/file
+              sudo_prefix: 'mysudo -c'
```
dankilman: consider adding an example with `sudo_prefix` and `use_sudo` combined, as putting this one here, doesn't actually do anything

content/plugins/fabric.md
```diff
@@ -153,6 +159,7 @@ node_templates:
             # Path to the script relative to the blueprint directory
             script_path: scripts/start.sh
             MY_ENV_VAR: some-value
+            use_sudo: false
```
dankilman: the basic example only specified the required inputs (+ some env var),
consider removing this and adding an example where this is set to true so that people don't confuse this for a required property (I know it is documented that it defaults to `false` but you know how it is with documentation. copy/paste is much more likely than reading the docs

content/plugins/fabric.md
```diff
@@ -58,6 +63,7 @@ node_templates:
                 - echo "source ~/myfile" >> ~/.bashrc
                 - apt-get install -y python-dev git
                 - pip install my_module
+              use_sudo: true
```
dankilman: perhaps add a comment before the use_sudo: true, specifying that if the property is left out, it is equivalent to passing false

content/plugins/fabric.md
```diff
@@ -206,8 +207,40 @@ node_templates:
...
+{{< /gsHighlight >}}
+
+
 # SSH configuration
-The fabric plugin will extract the correct host IP address based on the node's host. It will also use the username and key file path if they were set globally during the bootstrap process. However, it is possible to override these values and additional SSH configuration by passing `fabric_env` to operation inputs. This applies to `run_commands`, `run_task` and `run_module_task`. The `fabric_env` input is passed as is to the underlying [Fabric]({{< field "fabric_link" >}}/en/latest/usage/env.html) library, so check their documentation for additional details.
+
+The fabric plugin will extract the correct host IP address based on the node's host. It will also use the username and key file path if they were set globally during the bootstrap process. However, it is possible to override these values and additional SSH configuration by passing `fabric_env` to operation inputs. This applies to `run_commands`, `run_task` and `run_module_task`. The `fabric_env` input is passed as is to the underlying [Fabric]({{< field "fabric_link" >}}/en/1.8/usage/env.html) library, so check their documentation for additional details.
```
nir0s: replace version `latest` of the link with `1.8` since that's the version of fabric we're using.

PR #182, opened 2016-05-30
Update overview.md

zz19601:
According to the resource file, here should be 'node_instance_id' but not 'node_id'


idanmo:
Cherry picked to 3.4.0-build

PR #185, opened 2016-06-07
Inputs can be imported in DSL 1.3+

isaac-s:


nir0s:
node templates and outputs can also be imported. Please update.

isaac-s:
@nir0s done.

PR #192, opened 2016-06-14
Update fabric.md, valid schema section 'default' was added.

lexleogryfon:
'default:' section added for tasks_file, task_name, fabric_env to work with current cfy cli 3.3.1 , without 'default:' section blueprint cannot be validated & uploaded.

### DIFF
--- a/content/plugins/fabric.md
+++ b/content/plugins/fabric.md
@@ -77,8 +77,10 @@ node_templates:
         start:
           implementation: fabric.fabric_plugin.tasks.run_task
           inputs:
-            tasks_file: my_tasks/tasks.py
-            task_name: install_nginx
+            tasks_file: 
+             default: my_tasks/tasks.py
+            task_name: 
+             default: install_nginx
             task_properties:
               important_prop1: very_important
               important_prop2: 300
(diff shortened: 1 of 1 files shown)


nir0s:
Defaults are only relevant in `node_types`, not `node_templates`. If you provide the blueprint you have, we can see where the problem originates

lexleogryfon:
@nir0s
here is my blueprint https://github.com/lexleogryfon/cfy-examples-lex/tree/1184e80e679979069c7ac92a4d2aaab86a1a8fb9/apache-blueprint-for-manager
AFAIK, scripts & cloudify.interfaces.lifecycle should be defined in types.yaml, that is the correct way. So, I believe we have to change node_templates to node_types too.

lexleogryfon:
@nir0s please decide and choose one option,
1) right place to define fabric scripts is in node_types. And merge this pull request.
2) right place to define fabric scrips is in node_templates. Decline and close this pull request.
3) both places are correct.  Decline and close this pull request.

nir0s:
There is no "practice" for putting lifecycle.operations in `node_types` or `node_templates`. It depends on what you need. Specifically, in this situation, the lifecycle operations are defined under `node_templates` which means that they don't require the `default` key.

PR #194, opened 2016-06-16
CFY-5448 document powershell command_prefix auto deduction

nirfuchs:
updade script.md

using cloudify's script plugin with Windows PowerShell script
no longer requires to explicitly set 'command_prefix' to 'powershell'


content/plugins/script.md
```diff
@@ -239,13 +242,9 @@ node_templates:
       cloudify.interfaces.lifecycle:
         start:
           implementation: scripts/start.ps1
```
dankilman: use shorter form

PR #195, opened 2016-06-18
CFY-5471 manager upgrade docs

adamlavie:


content/manager/manager-upgrade.md
```diff
@@ -0,0 +1,219 @@
...
+Starting the upgrade process requires having the Cloudify Manager in [maintenance-mode](http://??). This will prevent having running executions during the upgrade process.
+To activate maintenance-mode, run:
+{{< gsHighlight  bash  >}}
+cfy maintenance-mode activate
+{{< /gsHighlight >}}
+The Cloudify Manager upgrade will be possible only once the manager enters maintenance mode.
+Once in maintenance-mode, using the pre-configured [simple-manager-blueprint-inputs.yaml](https://github.com/cloudify-cosmo/cloudify-manager-blueprints/blob/3.3.1/simple-manager-blueprint-inputs.yaml)
+execute the following CLI command to initiate the upgrade process:
+{{< gsHighlight  bash  >}}
+cfy upgrade --blueprint-path simple-manager-blueprint.yaml --inputs simple-manager-blueprint-inputs.yaml -v
```
funkyHat: Should this suggest `--install-plugins` as that's almost definitely going to be required on an upgrade?

content/manager/manager-upgrade.md
```diff
@@ -0,0 +1,219 @@
+---
```
nir0s: before reviewing, please change the name of the file to `upgrade.md`. We're already in the context of `manager`

And please also change the title to `Upgrading to a newer version`.

content/manager/manager-upgrade.md
adamlavie: Noted and added. Thanks!

PR #205, opened 2016-06-27
blueprint example

maimon33:


nir0s:
please replace all bash notes with markdown

PR #226, opened 2016-07-07
Fix the issue to no download get-cloudify.py from amazonaws.com #CFY-5688

yuexiao-wang:
Don't download get-cloudify.py from amazonaws.com,  so change the download page

### DIFF
--- a/content/intro/installation.md
+++ b/content/intro/installation.md
@@ -74,7 +74,7 @@ Run `python get-cloudify.py -h` for additional information.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


nir0s:
Why is downloading from amazon a problem?

yuexiao-wang:
Hi nir0s,
I can't access amazon to download the file  because of the corporate information security policy in China . As far as know, it is the same problem in many company in China. So I suggest change download page.

Please check CFY-5688 in JIRA.

Thanks!

nir0s:
As per https://cloudifysource.atlassian.net/browse/CFY-5688

PR #251, opened 2016-07-22
Modify version for dsl and types in blueprints.md

yuexiao-wang:
Modify version for dsl and types. in blueprints.md. It should be cloudify_dsl_1_3 for tosca_definitions_version and  import http://www.getcloudify.org/spec/cloudify/3.4/types.yaml in v3.5.0


idanmo:
Cherry picked to 3.4.0-build and 3.4.1-build.

PR #258, opened 2016-08-01
add examples for the new cli

maimon33:


content/cli/overview.md
```diff
@@ -9,55 +9,67 @@ weight: 1
 
 Cloudify's Command-Line Interface is the default method for interacting with Cloudify and managing your applications. It allows you to execute workflows on your local machine as well as interact with a running [Cloudify Manager]({{< relref "manager/getting-started.md" >}}) (to ssh into a running Manager, upload blueprints, delete them, create deployments, execute workflows, retrieve events and more).
 
+The Cloudify Command-Line has two working modes: Local and Manager.
```
nir0s: The CLI has two working modes... `local` and `manager`

content/cli/overview.md
```diff
@@ -9,55 +9,67 @@ weight: 1
...
+Local will be used to run functions (eg. deployments) for the machine you are working on. Manager will use a Cloudify manager server
```
nir0s: Working locally means running workflows directly from the machine the CLI is installed on. Working with a manager means executing workflows from a manager instead.

Using the CLI to control a manager adds additional functionality to the CLI which is not exposed in `local` mode such as ssh-ing into a running manager, downloading its logs, creating snapshots, uploading plugins and more...

NOTE:
Running commands not supposed in local mode will show an error stating that they're only supported when using a manager. To use a manager, you can either run `cfy bootstrap` to bootstrap a new manager or `cfy use` to use an existing one. LINKS TO USE AND BOOTSTRAP

PR #274, opened 2016-08-25
CFY-5871: Installing from GitHub requires git client

isaac-s:


chanagershom:
@isaac-s, I am getting a message that this branch has conflicts that need to be resolved, but I don't see them? Are you able to see/resolve so that I can merge this?

isaac-s:
@chanagershom done. However remember that this is for 3.4.0... we need to replicate this change on the 4.0.0. Can you do this, or do you need a separate PR?

chanagershom:
@isaac-s  It would be helpful, this time round, if you could do it with a separate PR for 4.0. Thanks!

isaac-s:
@chanagershom just checked and the 4.0.0 version of the file is already correct.

PR #296, opened 2016-12-08
Add get_attribute restriction

Jessiebr:


content/blueprints/spec-intrinsic-functions.md
```diff
@@ -348,6 +348,7 @@ This is because they all implicitly belong to the same scaling group instance (t
 * If an attribute is not found in the inspected node instance runtime properties, the scan will fall back to the matching node properties. If the attribute is not found in the node properties as well, `null` is returned.
 * `SELF` can only be used in interface operation inputs.
 * `SOURCE` and `TARGET` can only be used in relationship interface operation inputs.
+* Node properties are immutable, therefore are set before runtime. `get_attribute` is used to set runtime properties, therefore can't be used on node properties.
```
EarthmanT: Node Properties are immutable, and are set before runtime. Since get_attribute retrieves runtime_properties which are set during runtime, it cannot be used to set node properties.

PR #301, opened 2016-12-30
OPENSTACK-41 support for Liberty & Mitaka

funkyHat:
Add version compatibility & note about keystone URL

PR #315, opened 2017-01-24
Added Mitaka support description

sivanbar:


funkyHat:
Could you check #301 please? (It's been open a while but I should have chased it up...)

PR #319, opened 2017-01-26
Add vsphere plugin 2.2.0 documentation

geokala:


content/plugins/vsphere.md
```diff
@@ -126,6 +126,10 @@ Each type has property `connection_config`. It can be used to pass parameters fo
 
 * `windows_timezone` The timezone to set the Windows system to. It will default to 90 (GMT without daylight savings (approximately UTC)). If you wish to set this it must be set to an [appropriate integer value](https://msdn.microsoft.com/en-us/library/ms912391%28v=winembedded.11%29.aspx)
 
+* `windows_organization` The organization name to set on the Windows system. It will default to: Organization
+
+* `custom_sysprep` A custom sysprep answers file to use for full customization of Windows. Please note that this should be verified to work beforehand, as any errors will only appear on Windows and will not be visible to the plugin. Please note also that any scripts, etc, which attempt to work on the VM after the custom sysprep must be tolerant of multiple retries, as the plugin cannot detect when the custom sysprep has finished so will provide the server as soon as IPs are assigned (which will be before customization is complete).
```
chanagershom: A custom ~sysprep~ **System Preparation Answers**  ~answers~ file to use for full customization of Windows. ~Please n~  Note that this **file** should be verified to work **correctly** before ~hand~ **being applied**, as any errors will ~only~ appear **only** on Windows and will not be visible to the plugin.~Please note also~ **Also note** that any scripts, etc, ~which~ **that** attempt to work on the VM after the custom sysprep must ~be tolerant~ **tolerate** ~of~ multiple retries ~, as~ **because the plugin cannot detect when the custom sysprep has finished **,** so ~will~ provide**s** the server as soon as **the** IPs are assigned (which ~will be~ **occurs** before customization is complete).

PR #321, opened 2017-01-31
New clarify without comments

EarthmanT:


funkyHat:
gcp.md seems to be in the wrong place? (I mentioned this in the other PR but I guess that was missed)

PR #326, opened 2017-02-07
fix line endings

funkyHat:
Looks like this file was accidentally saved in old-MacOS mode, or git has "helpfully" configured itself to do some line-ending translation.


funkyHat:
Was fixed separately

PR #335, opened 2017-02-13
Adding clustering docs init

chanagershom:
Please take a look at the high-availability-clusters file and let me know if there is anything else that needs to be added. Thanks!


content/manager/high-availability-clusters.md
```diff
@@ -0,0 +1 @@
...
##Failure of the Master Cloudify Manager
As already mentioned, in the event that the active Cloudify Manager fails, one of the hot standbys is activated. It is important that the user investigates and fixes the issues that caused the original master to fail so that high availability is maintained and to avoid having a single point of failure.

##Important Considerations
Be aware of the following considerations when you are implementing clustering.

* If the active Cloudify Manager was processing a workflow at the time it fails, the new active manager does not resume and complete the workflow.
* Policies are not synchronized between Cloudify Managers in the cluster.
* If there is a loss of connection between the Cloudify Managers in the cluster, they independently assume the `active` role. When the connection is resumed, the most recently updated Cloudify Manager becomes the master. Data that was accumulated on the other Cloudify Managers during the disconnection is not synchronized, so is lost. 
* A Cloudify Manager that is down remains in the cluster unless you remove it. To remove a Cloudify Manager, run `cluster-remove-node`.
```
sivanbar: 1. Enterprise version -> rename to Premium version
2. I'd remove the first disclaimer about 3 managers in a cluster, as it is described in the details section
3. You just specify `cluster-join` when you bring the Cloudify Manager up - you first specify cluster-start to start a cluster, then cluster-join from any manager that joins the cluster - so I would either replace the Cloudify manager with "a Cloudify manager" or add this clarification
4. Creating a cluster - can be done at any time post bootstrap, even after there is data in the manager
5. cluster-join - command should be triggered post bootstrap. The joining manager has to be empty or the command fails
6. State is linked - worth noting this is the data kept by the manager, not the manager's state (running, stopped, etc)
7. I'd change the sentence with console to "Cloudify automatically detects the active manager, using Console"
8. "However, the you do need to know which Cloudify Manager nodes are in the cluster" - not sure I understand this one... You don't actually, as a user, have to know this :)
9. "It is important that the user investigates and fixes the issues that caused the original master to fail" - or alternatively add another manager to the cluster. Note that since operations cannot be performed on the non active manager, troubleshooting has to be done by SSHing the manager.

PR #339, opened 2017-02-16
Vsphere 2.3.0

funkyHat:


funkyHat:
Plugin link will be broken if this is merged before cloudify-vsphere-plugin 2.3 is released

PR #347, opened 2017-03-04
Update sudo info

isaac-s:
Using `sudo` is mandatory if bootstrapping using a non-root user and using `--install-plugins`.

### DIFF
--- a/content/manager/bootstrapping.md
+++ b/content/manager/bootstrapping.md
@@ -22,7 +22,6 @@ See the reference for bootstrapping on [Openstack]({{< relref "manager/bootstrap
 
 To bootstrap a Cloudify Manager:
 
-
 # Initialize a Working Directory
 
 Navigate to a directory of your choosing, and initialize it as a Cloudify working directory using this command:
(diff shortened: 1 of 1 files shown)


content/manager/bootstrapping.md
```diff
@@ -147,17 +146,19 @@ After downloading the manager resources package, and placing it in an accessible
...
 {{< gsHighlight  sh  >}}
 $ cfy bootstrap --install-plugins -p /path/to/manager/blueprint/file -i /path/to/inputs/yaml/file
 ...
-
 {{< /gsHighlight >}}
 
+{{% gsInfo title="Info" %}}
+If you are using the CLI environment that has been installed by the official CLI RPM, and are using the `--install-plugins` parameter, you will have to execute the `bootstrap` command using `sudo`. That is because the Cloudify Python virtual environment (located in `/opt/cfy/embedded`) is owned by `root`. Alternatively, you could change the ownership of `/opt/cfy/embedded` (and its descendants) so the user running the bootstrap has write-level access.
+{{% /gsInfo %}}
+
```
chanagershom: If you are using ~the~ a CLI environment that ~has been~ **was** installed ~by~ **using** the official CLI RPM, and are using the `--install-plugins` parameter, you ~will have to~ **must** execute the `bootstrap` command using `sudo` ~. That is~ because the Cloudify Python virtual environment (located in `/opt/cfy/embedded`) is owned by `root`. Alternatively, you~ could~ **can** change the ownership of `/opt/cfy/embedded` (and its descendants) so the user running the bootstrap has write-level access.

isaac-s:
Thanks @chanagershom. I fixed the text. The typo in my new commit comment was completely accidental.,,

PR #379, opened 2017-04-30
Updated ports

isaac-s:
8101 isn't used anymore.
5672 isn't used anymore — we're working with SSL now, so 5671 is used instead.


chanagershom:
Thanks Isaac. This page had been checked and approved, but clearly not thoroughly enough.

chanagershom:
!

PR #380, opened 2017-04-30
Added fixes to secrets cli

inbala:


content/cli/secrets.md
```diff
@@ -21,22 +21,24 @@ These will work on each command:
...
 Create a new secret (key-value pair)
 
 `KEY` is the new secret's key
 
 #### Required flags
 
-*  `-s, --secret_value TEXT` - The secret's value to be set. 
+One of these flags :
+*  `-s, --secret-string TEXT` - The string to use as the secret's value.
+*  `-f, --secret-file TEXT` - The secret's file to use its content as value to be set.
```
chanagershom: Firstly, is this the _name_ of the secret file? If so, replace with the following text.
```The name of the secret file that contains the values to be set.```

I don't think that we need to say anything about "content" because the fact that we refer to values implies that that is the content.

PR #423, opened 2017-07-16
update upgrade doc

dusking:


chanagershom:
This merge is waiting until the "patch-1" has been created and is available. Check with @dusking before merging.

PR #428, opened 2017-07-20
updating config in OpenStack PIugin

chanagershom:
Please review Configuration and Accessing Secrets sections.


funkyHat:
Looks like I've got here a bit too late, but these new secret store sections don't make sense. Each plugin has a different set of parameters which need to be provided. After merging this PR all plugin pages describe the way that will only work for OpenStack

chanagershom:
We are customizing each the page for each plugin. OpenStack is just the one
that we started with.

Chana Gershom, Technical Writer <http://cloudify.co>
------------------------------
M: +972547718415 http://cloudify.co @cloudifysource
<https://twitter.com/CloudifySource>
<https://www.linkedin.com/company-beta/17918192/>
<https://github.com/cloudify-cosmo>
<https://www.youtube.com/cloudifysource>

PR #434, opened 2017-07-26
(temporarily) remove secrets section from GCP page

funkyHat:
Should be added with updated keys once https://cloudifysource.atlassian.net/browse/GCP-41 is done


chanagershom:
Approved as a temporary measure.

PR #438, opened 2017-08-02
Update plugins' configuration

chanagershom:
This PR relates to updates relating to moving the configuration information higher on the page for each plugin, and providing the necessary information for working with the secrets store. In addition, general updates, relating to formatting has also been applied. Happy reading!


content/plugins/gcp.md
```diff
@@ -278,7 +304,7 @@ mx:
 
 {{< /gsHighlight >}}
 
-The DNSRecord type can be connected to an instance or directly to an IP. In this case the (associated) public IP will be added to the list of resources.
+The DNSRecord type can be connected to an instance, or directly to an IP address. In the case of an IP address, the (associated) public IP is added to the list of resources.
```
funkyHat: This is less correct than before. In *either* the `IPAddress` case *or* the `Instance` case the IP address (of the IPAddress or Instance) will be added to the list of resources (I agree the previous text was only *technically* more correct, while being more confusing :).

Perhaps... "The DNSRecord The DNSRecord type can be connected to a `cloudify.gcp.nodes.Instance`, or to a `cloudify.gcp.nodes.IPAddress`. Each instance or IP address node that is connected will have its public IP address added to the list of resources for this record." (feel free to revise further for clarity :)

funkyHat: (revised to add the relationships that must be used) "The DNSRecord The DNSRecord type can be connected to a `cloudify.gcp.nodes.Instance` (using `cloudify.gcp.relationships.dns_record_connected_to_instance`), or to a `cloudify.gcp.nodes.IPAddress` (using `cloudify.gcp.relationships.dns_record_connected_to_ip`). Each instance or IP address node that is connected will have its public IP address added to the list of resources for this record."

content/plugins/gcp.md
```diff
@@ -637,55 +663,55 @@ A GCP Instance (i.e. a VM).
 **Properties:**
 
 
-  * `scopes`
-    Optional scopes. If not will set by default:  'https://www.googleapis.com/auth/devstorage.read_write', 'https://www.googleapis.com/auth/logging.write'
+  * `scopes` - 
+    Optional scopes. If not, the type is derived from:  'https://www.googleapis.com/auth/devstorage.read_write', 'https://www.googleapis.com/auth/logging.write'
```
funkyHat: "default" is more correct than "derived from" here. It's just a list of strings not a type

content/plugins/vsphere.md
```diff
@@ -54,19 +52,34 @@ The vSphere plugin allows users to use a vSphere based infrastructure for deploy
...
     type: cloudify.vsphere.nodes.Network
     properties:
-      vsphere_config:  
-        host: { get_secret: vsphere_host }
-        port: { get_secret: vsphere_port }
-        username: { get_secret: vsphere_username }
-        password: { get_secret: vsphere_password }
+
+      vSphere_config:  
+        username: { get_secret: keystone_username }
```
funkyHat: These keys are for OpenStack again, not vSphere

content/plugins/vsphere.md
```diff
@@ -54,19 +52,34 @@ The vSphere plugin allows users to use a vSphere based infrastructure for deploy
...
+        auth_url: { get_secret: keystone_url }
+        region: { get_secret: region }
+
  {{< /gsHighlight >}}   
 
+## Providing Credentials as Environment Variables that are not Stored as Secrets
+If you do not use secret storage, you must provide the following credentials as environment variables:
+{{< gsHighlight  yaml  >}}
+       vSphere_config:  
+        username: { keystone_username }
```
funkyHat: Same as above

content/plugins/gcp.md
```diff
@@ -24,6 +24,39 @@ Be aware that some services and resources vary in availability between regions a
...
+        region: { get_secret: region }
+ {{< /gsHighlight >}}   
+
+
+### Providing Credentials as Environment Variables that are not Stored as Secrets
+If you do not use secret storage, you must provide the following credentials as environment variables:
+
+{{< gsHighlight  yaml  >}}
+       gcp_config:  
+        username: { keystone_username }
```
funkyHat: This doesn't make sense to me, unless there's some weird env var syntax I don't know about :)
Also talking about environment variables doesn't make much sense for manager based deployments.
These should probably be `{ get_input: ... }` instead

funkyHat: (though actually we should probably just remove this bit along with the secrets bit until GCP-41 is done)

content/plugins/vsphere.md
funkyHat: (also should probably be `{ get_input: ... }`)

content/plugins/gcp.md
chanagershom: Okay. I am going to remove secrets and env variables. I don't remember readding it. Maybe I hadn't updated my local branch.

content/plugins/gcp.md
chanagershom: done

content/plugins/diamond.md
```diff
@@ -82,87 +78,15 @@ node_templates:
...
 Two interfaces are involved in setting up a monitoring agent on a machine:
 
-* `cloudify.interfaces.monitoring_agent` - The interface in charge of installing, starting stopping and uninstalling the agent.
-* `cloudify.interfaces.monitoring` - The interface in charge of configuring the monitoring agent.
+* `cloudify.interfaces.monitoring_agent` - The interface that manages installing, starting, stopping, and uninstalling the agent.
+* `cloudify.interfaces.monitoring` - The interface that manages configuring the monitoring agent.
 
 The example above shows how the Diamond plugin maps to these interfaces.
 
-# Global config
```
funkyHat: Is this section no longer relevant/incorrect? We seem to be removing a big chunk without putting equivalent info back somewhere else?

content/plugins/host-pool.md
```diff
@@ -73,59 +70,58 @@ Base type for a pool host.
 
 **Properties:**
 
-  * `os` windows.
+  * `os` - The Windows operating system.
```
funkyHat: This is an overridden default for the `os` property (because `...WindowsHost` derives from `...Host`), and it is literally "windows". Given that this is a property that shouldn't be overridden ever, perhaps we should make that clear instead, e.g.
```markdown
* `os`: `windows`  (this is an implementation detail and should not be overridden)
```

content/plugins/host-pool.md
```diff
@@ -64,7 +61,7 @@ Base type for a pool host.
 
 **Properties:**
 
-  * `os` linux.
+  * `os` - The Linux operating system.
```
funkyHat: Same as `...WindowsHost` below

content/plugins/script.md
```diff
@@ -227,12 +223,12 @@ node_templates:
               command_prefix: /opt/ruby/bin/ruby
 {{< /gsHighlight >}}
 
-This will execute `start.rb` with the ruby binary in `/opt/ruby/bin/ruby`
+This executea `start.rb` with the ruby binary in `/opt/ruby/bin/ruby`.
```
funkyHat: executes

content/plugins/script.md
```diff
@@ -437,15 +433,15 @@ ctx.instance.runtime_properties['my_property'] = 'my_value'
 ctx.instance.runtime_properties['my_properties']['my_nested_property'] = 'nested_value'
 {{< /gsHighlight >}}
 
-Once a dict attribute is discovered during the attribute search the following logic applies:
+Once a `dict` attribute is discovered during the attribute search, the following logic applies:
```
funkyHat: dictionary?

chanagershom:
Not sure why this was removed. There probably was a good reason but at this
point I am unaware of it. I have readded it in the meantime.

PR #474, opened 2017-11-02
CFY-7463 Header Section

mxmrlv:


content/plugins/aria.md
```diff
@@ -3,13 +3,11 @@ layout: bt_wiki
...
 draft: false
-weight: 2
-6
-weight: 10000
+weight: 100
 ---
 {{% gsSummary %}} {{% /gsSummary %}}
 
-Plugin Description
+The ARIA plugin enables to use TOCSA-based service-templates for deploying services and applications.
```
aviyoop: The ARIA plugin enables to use TOCSA-based service templates for deploying services and applications

PR #516, opened 2017-11-21
Remove Grafana references

talisela:


talisela:
Remove Grafana reference

PR #542, opened 2017-12-24
Rename availability to visibility

inbala:
Also, add information about creating a resource with visibility and updating the visibility.


bmansheim:
Do you want to cancel this pull request and replace it with the new terminology?

PR #571, opened 2018-01-23
Update upgrade procedures

bmansheim:


content/installation/migrating-manager.md
```diff
@@ -0,0 +1,32 @@
+---
+layout: bt_wiki
+title: Reinstalling and Migrating Cloudify Manager
+category: Installation
+draft: true
+weight: 400
+
+---
+
+To move your Cloudify Manager to another host, you must use same process as a reinstall. One reason to migrate is to move your Cloudify Manager to a host with more resources. You can do this in one of these ways:
```
geokala: This file just seems to be a summarised rewrite of the actual snapshot restore procedure. It might be better if it just stated that migrating a manager is done by following the snapshots procedure, and linked to it? Otherwise there's a high chance that the two different lists of steps will become out of sync and cause confusion.

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
-
-* Reinstall -
-
-  1. Save snapshot of the Cloudify Manager.
-  1. Uninstall the Cloudify Manager from the host
-  1. Install the Cloudify Manager on the host.
-  1. Restore snapshot of the Cloudify Manager to the host.
-
-* Migration -
+{{% gsWarning title="Downtime" %}}
```
geokala: We seem to be summarising the process in a different place again (see comment on migrations).

content/cli/snapshots.md
```diff
@@ -11,7 +11,7 @@ The `cfy snapshots` command is used to manage data snapshots of Cloudify manager
 
 You can use the command to create, upload, download, delete and list snapshots and also to restore a Manager using a snapshot archive.
 
-See [snapshots]({{< relref "manager/snapshots.md" >}}) for more information.
+For more about working with snapshots, go to: [snapshots]({{< relref "manager/snapshots.md" >}}).
```
geokala: This has the manager/snapshots.md path, but you closed the PR which actually made that file accurate so there's no point linking to the broken one.

content/installation/migrating-manager.md
```diff
@@ -0,0 +1,32 @@
...
+
+* **In-place reinstall** - Replace Cloudify Manager with the same version on the same host
+* **Migration reinstall** - Replace Cloudify Manager with the same version on another host
+
+The reinstall process includes:
+
+* In-place reinstall -
+
+  1. Backup a snapshot of the Cloudify Manager.
+  1. Backup the agents certificates from the Cloudify Manager.
```
geokala: I don't believe this step is required- where does it come from?

Unless you mean copying all of the certificates, which is only applicable for a Cloudify 4.0.0 manager on an in-place upgrade.

geokala: You also missed copying the rest-security.conf from managers prior to 4.2.0.

content/installation/migrating-manager.md
```diff
@@ -0,0 +1,32 @@
...
+  1. Uninstall the Cloudify Manager from the host
+  1. Install the Cloudify Manager on the host.
+  1. Restore the snapshot of the Cloudify Manager to the host.
```
geokala: At this stage, restoring the rest-security.conf might be required (if it needed to be backed up, above).

geokala: Also, it might be necessary to update the agents here (cfy agents install --all-tenants)

content/installation/migrating-manager.md
```diff
@@ -0,0 +1,32 @@
...
+
+* Migration reinstall -
+
+  1. Backup a snapshot of the Cloudify Manager.
+  1. Backup the agents certificates from the Cloudify Manager.
```
geokala: What is the purpose of backing up the agent certs here? The only place I am aware of it being needed is a 4.0.0 manager on an in-place upgrade, and this is referring to a migration.

content/installation/migrating-manager.md
```diff
@@ -0,0 +1,32 @@
...
+  1. Install the Cloudify Manager on the host.
+  1. Restore the snapshot of the Cloudify Manager to the host.
+  1. (Optional) Uninstall the Cloudify Manager from the original host. In Cloudify 4.2 and below, this is called teardown.
```
geokala: This step MUST be done after the agents install. If it is done at this stage it'll be an expensive support outing to fix every single agent in the customer's environment.

content/installation/migrating-manager.md
```diff
@@ -0,0 +1,32 @@
...
+  1. Migrate agents from the old Cloudify Manager.
```
geokala: I would recommend after this stage that the old manager be shut down, even if it is not to be deleted. If any auto-heal policies are in use then the old manager MUST also be disconnected from any networks on which it could reach the platforms it is managing. Otherwise, after the first auto-heal occurs, you will end up with the old manager auto-healing as well, and thus get two copies of compute nodes for a deployment, each orphaned from one of your managers.

content/installation/migrating-manager.md
```diff
@@ -0,0 +1,32 @@
...
+  1. Uninstall the Cloudify Manager from the host
+  1. Install the Cloudify Manager on the host.
+  1. Restore the snapshot of the Cloudify Manager to the host.
+
+* Migration reinstall -
+
+  1. Backup a snapshot of the Cloudify Manager.
+  1. Backup the agents certificates from the Cloudify Manager.
+  1. Install the Cloudify Manager on the host.
+  1. Restore the snapshot of the Cloudify Manager to the host.
```
geokala: On a 4.2 snapshot, you have to wait after the snapshot restore is complete for the rest service to restart. Recommended is at least 10 seconds.

content/installation/migrating-manager.md
geokala: On a 4.2 snapshot, you have to wait after the snapshot restore is complete for the rest service to restart. Recommended is at least 10 seconds.

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
 
-* **Reinstall** - Replace Cloudify Manager with the same version on the same host
-* **Migration** - Replace Cloudify Manager with the same version on another host
 * **In-place upgrade** - Replace Cloudify Manager with a higher version on the same host
 * **Migration upgrade** - Replace Cloudify Manager with a higher version on another host
 
 {{% gsNote title="Version Relevance" %}}
-The upgrade process is supported for upgrade from any currently supported version. To upgrade from an unsupported version, contact Cloudify Support.
-{{% /gsNote %}}
+The upgrade process is supported for upgrade from any currently supported version to the latest GA version. To upgrade from an unsupported version, contact Cloudify Support.
```
geokala: It would be good if this linked to somewhere listing the supported versions.

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
-      ```curl -o ~/cfy_teardown_4_0_0.sh https://raw.githubusercontent.com/cloudify-cosmo/cloudify-dev/master/scripts/cfy_teardown_4_0_0.sh```<br>
+## Backup Agent Certificates
 
-      ```sudo bash cfy_teardown_4_0_0.sh```. You must supply an -f flag.<br>
-      For high availability clusters, you must also run ```https://github.com/cloudify-cosmo/cloudify-dev/blob/master/scripts/delete_cluster_4_0_1.py```.
+To backup agent certificates:
 
-1. (Optional) We recommend that you run this command to remove the profile directory of this Cloudify Manager from your local `~/.cloudify/profiles` directory:
-   ```rm -rf ~/.cloudify/profiles/{{ your Manager’s IP address }}```
+* Backup the `/etc/cloudify/ssl` directory to a remote host.
```
geokala: This is only necessary on an in-place 4.0.0 upgrade to the best of my knowledge.

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
+The upgrade process includes:
 
 * In-place upgrade -
 
-  1. Save snapshot of Cloudify Manager.
-  1. Uninstall (teardown in Cloudify 4.2 and below) the Cloudify Manager from the host.
-  1. Install the Cloudify Manager latest version on the host.
-  1. Restore snapshot of the Cloudify Manager to the host.
+  1. [Backup a snapshot]({{< relref "installation/upgrading-manager.md#backup-a-snapshot" >}}) of the Cloudify Manager.
+  1. [Backup the agents certificates]({{< relref "installation/upgrading-manager.md#backup-agent-certificates" >}}) from the Cloudify Manager.
```
geokala: I think this step is probably not necessary unless performing an in-place upgrade from 4.0.0.

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
-  1. Migrate agents from the old Cloudify Manager.
+The upgrade process includes:
 
 * In-place upgrade -
 
-  1. Save snapshot of Cloudify Manager.
-  1. Uninstall (teardown in Cloudify 4.2 and below) the Cloudify Manager from the host.
-  1. Install the Cloudify Manager latest version on the host.
-  1. Restore snapshot of the Cloudify Manager to the host.
+  1. [Backup a snapshot]({{< relref "installation/upgrading-manager.md#backup-a-snapshot" >}}) of the Cloudify Manager.
```
geokala: You must also copy the rest-security.conf if you are upgrading from a version prior to 4.2.0 and using multiple cloudify users.

content/installation/upgrading-manager.md
geokala: (also worth noting that the certificates in the directory you refer to are not just agent certificates, it's all the manager's ssl certs)

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
+  1. [Backup the agents certificates]({{< relref "installation/upgrading-manager.md#backup-agent-certificates" >}}) from the Cloudify Manager.
+  1. [Uninstall the Cloudify Manager]({{< relref "installation/upgrading-manager.md#uninstall-cloudify-manager" >}}) from the original host. In Cloudify 4.2 and below, this is called teardown.
+  1. [Install the Cloudify Manager]({{< relref "installation/upgrading-manager.md#install-cloudify-manager" >}}) latest version on the host.
+  1. [Restore the snapshot]({{< relref "installation/upgrading-manager.md#restore-snapshot" >}}) of the Cloudify Manager to the host.
```
geokala: At this stage, the rest-security.conf might need restoring too, if you were upgrading from a pre 4.2.0

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
 
 * Migration upgrade -
 
-  1. Save snapshot of the the Cloudify Manager.
-  1. Install the Cloudify Manager on the host.
-  1. Restore snapshot of the Cloudify Manager to the host.
-  1. (Optional) Uninstall (teardown in Cloudify 4.2 and below) the Cloudify Manager from the host.
-  1. Migrate agents from the old Cloudify Manager.
+  1. [Backup a snapshot]({{< relref "installation/upgrading-manager.md#backup-a-snapshot" >}}) of the Cloudify Manager.
+  1. [Backup the agents certificates]({{< relref "installation/upgrading-manager.md#backup-agent-certificates" >}}) from the Cloudify Manager.
```
geokala: Not required here as far as I'm aware.

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
+  1. [Install the Cloudify Manager]({{< relref "installation/upgrading-manager.md#install-cloudify-manager" >}}) latest version on the host.
+  1. [Restore the snapshot]({{< relref "installation/upgrading-manager.md#restore-snapshot" >}}) of the Cloudify Manager to the host.
+  1. (Optional) [Uninstall the Cloudify Manager]({{< relref "installation/upgrading-manager.md#uninstall-cloudify-manager" >}}) from the original host. In Cloudify 4.2 and below, this is called teardown.
```
geokala: This step MUST come after the agents install.

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
+  1. [Migrate agents]({{< relref "installation/upgrading-manager.md#migrate-agents" >}}) to the new Cloudify Manager.
```
geokala: After the agents install, the old manager should be powered off (whether or not it is deleted), and if auto-healing policies are used then it MUST be disconnected from all networks that could allow it to conduct healing workflows.

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
-   ```cfy snapshots upload {{ /path/to/the/snapshot/file }} --snapshot-id my_snapshot```<br>
+   ```cfy snapshots upload {{ /path/to/the/snapshot/file }} --snapshot-id my_snapshot```
 
-1. To restore the snapshot to Cloudify 4.x and above, run: <br>
+1. To restore the snapshot to Cloudify 4.x and above, run:
 
   * For Cloudify Manager 4.0.1 and above: ```cfy snapshots restore my_snapshot --restore-certificates```
   * For Cloudify Manager 4.0.0: ```cfy snapshots restore my_snapshot```
 
 1. To check the status of the execution after it is complete, run:
```
geokala: And after you do this if you're restoring from pre-4.2 and using multiple users then you need to restore the rest-security.conf.

If you're restoring a 4.2+ snapshot then the rest service will restart shortly after the execution finishes, so you should wait at least 10 seconds (new users will be usable after it restarts).

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
    ```cfy executions list --include-system-workflows```
 
+## Migrate Agents
+
 1. (Optional) To apply the agent certificates from the previous Manager, run these commands to replace the new SSL directory with the copied one.
```
geokala: This must NOT be done unless the certificates are referring to the same IPs/hostnames that are relevant for the new manager as well as the old one. Otherwise the agents simply can't work.
These steps should probably be changed to include taking a copy of the SSL certs dir in case someone doesn't know exactly what they're doing but blindly follows them.

Also, this agent cert restore should only be required on an in-place upgrade from 4.0.0

content/installation/upgrading-manager.md
```diff
@@ -101,12 +118,11 @@ We recommend that premium users contact Cloudify Support for additional assistan
 
 1. Reboot the virtual instance with the new Cloudify Manager.
 
-1. If you have running agents, make sure that you apply `patch-1`, then run `cfy agents install`.
-
+1. If you have running agents, run `cfy agents install`.
```
geokala: --all-tenants

Otherwise you'll get a nasty surprise later.

content/installation/migrating-manager.md
bmansheim: So I should change step 2 from:
Incorrect: Backup the agents certificates from the Cloudify Manager.
Correct: (For upgrade from 4.1.x and below) Backup the rest-security.conf file to a separate location.
Also, what's the path to rest-security.conf?

content/installation/migrating-manager.md
bmansheim: Because this is only for reinstall of 4.3, I'll skip the rest-security.conf issue. Otherwise, I'll add: Restore the rest-security.conf file to <path>.
Should I add cfy agents install --all-tenants? If you don't know, who does?
Should I also write this for a 4.3 snapshot? When the snapshot restore is complete, the REST service restarts. (Duration > 10 secs)

content/installation/migrating-manager.md
bmansheim: So you don't need to backup agent certs to migrate a 4.3 manager to another host?

content/installation/migrating-manager.md
bmansheim: I'll add a warning: If you do not uninstall the old Cloudify Manager and your workflows include auto-heal workflows, you must disconnect it from any networks that it can use to communicate with managed hosts.

content/installation/upgrading-manager.md
bmansheim: I'll add a warning: If you do not uninstall the old Cloudify Manager and your workflows include auto-heal workflows, you must disconnect it from any networks that it can use to communicate with managed hosts.

content/installation/upgrading-manager.md
bmansheim: Based on your migration comments:
1. I'll preface step 2 with: (For upgrade from Cloudify Manager 4.0.0)
2. I'll add after step 2: (For upgrade from Cloudify Manager 4.1.x and below) Backup the rest-security.conf file to a separate location.

content/installation/upgrading-manager.md
bmansheim: I'll add:
(For upgrade from Cloudify Manager 4.1.x and below) Restore the rest-security.conf file to ____.
(For upgrade from Cloudify Manager 4.2 and above) When the snapshot restore is complete, the REST service restarts. (Duration > 10 secs)

content/installation/upgrading-manager.md
bmansheim: I'll add: (For in-place upgrade from Cloudify Manager 4.0.0)
Where is the ssl certs dir?

content/installation/migrating-manager.md
geokala: If this is purely for a 4.3 reinstall then neither of those steps are needed.

Doesn't this mean that we're duplicating a bunch of steps though, by having a page specifically for 4.3 and others for others?

geokala: Also, if this is only for 4.3 to 4.3 reinstall I think a caution note is warranted pointing someone at the migration page because many of the steps are similar so it could lead to harmful confusion.

content/installation/migrating-manager.md
geokala: cfy agents install --all-tenants is correct, though if you're doing in place install I think it probably isn't required- this needs testing.

geokala: And yes, for the 4.3 restore it'll restart the rest service so you need to wait for ~10 seconds.

content/installation/migrating-manager.md
geokala: Correct- only 4.0.0 required the agent certs to be backed up I think- we've been including them in the snapshot since then.

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,113 @@
...
-  1. Install the Cloudify Manager on the host.
-  1. Restore snapshot of the Cloudify Manager to the host.
-  1. (Optional) Uninstall (teardown in Cloudify 4.2 and below) the Cloudify Manager from the host.
-  1. Migrate agents from the old Cloudify Manager.
+  1. [Backup a snapshot]({{< relref "installation/upgrading-manager.md#backup-a-snapshot" >}}) of the Cloudify Manager.
+  1. [Backup the agents certificates]({{< relref "installation/upgrading-manager.md#backup-agent-certificates" >}}) from the Cloudify Manager.
+  1. [Install the Cloudify Manager]({{< relref "installation/upgrading-manager.md#install-cloudify-manager" >}}) latest version on the host.
+  1. [Restore the snapshot]({{< relref "installation/upgrading-manager.md#restore-snapshot" >}}) of the Cloudify Manager to the host.
+  1. [Migrate agents]({{< relref "installation/upgrading-manager.md#migrate-agents" >}}) to the new Cloudify Manager.
+  1. (Optional) [Uninstall the Cloudify Manager]({{< relref "installation/upgrading-manager.md#uninstall-cloudify-manager" >}}) from the original host. In Cloudify 4.2 and below, this is called teardown.
```
geokala: We should probably still have a note here about making sure it's disconnected from the platform if it's not destroyed.

content/installation/upgrading-manager.md
geokala: (because the agents will still be connected to the old manager, and when you run cfy_agents install --all-tenants they'll be pointed to the new manager using the old manager to pass them tasks)

(also, the certificate will be wrong for the new manager unless it has the same IP address and/or hostname that the agents use to refer to the manager)

content/installation/upgrading-manager.md
geokala: It's more involved than that- once you've restored the rest-security.conf you have to then restore the admin user from the snapshot into the database. This is probably not something that we can reliably quickly document so I suggest we leave it as a 'contact support' note. Anyone not wanting to ask for support on it is able to just reset the user passwords.

4.2+ though, yes, the wait is required.

content/installation/upgrading-manager.md
geokala: /etc/cloudify/ssl- it's the one referenced in the commands

content/installation/migrating-manager.md
bmansheim: Thanks. Keep in mind that once we have all of the correct information for all 4.3 scenarios, I will reuse sections in the different places so it the same steps in each process are updated together.

bmansheim:
@geokala I didn't understand the last comment but there is no field there for me to write back. You want me to add the text in parentheses to the backup agent certificates instruction?

content/installation/upgrading-manager.md
bmansheim: So what should be backed up for just the agent certs?

content/installation/upgrading-manager.md
```diff
@@ -2,96 +2,134 @@
...
+## Restore User Database
+
+For upgrades from a pre-4.2 snapshot and with multiple users, either:
+
+* Reset the passwords of all old users
+* Contact support to restore the old users with the `/opt/manager/rest-security.conf` file.
+
+## Migrate Agents
+
 1. (Optional) To apply the agent certificates from the previous Manager, run these commands to replace the new SSL directory with the copied one.
```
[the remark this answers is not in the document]
bmansheim: 1. For in-place upgrade from Cloudify Manager 4.0.0 only:
    1. To apply the agent certificates from the previous Manager, run these commands to replace the new SSL directory with the copied one.
      ```sudo rm -rf /etc/cloudify/ssl```<br>
      ```sudo cp -r {{ the previously saved SSL directory. For example, `/home/centos/ssl` }} /etc/cloudify```
    1. To ensure that the directory has Read permissions, run:
      ```sudo chmod -R 755 /etc/cloudify/ssl```
    1. Reboot the virtual instance with the new Cloudify Manager.
1. If you have running agents, run `cfy agents install --all-tenants`.

content/installation/upgrading-manager.md
bmansheim: @geokala Please reply

content/installation/upgrading-manager.md
```diff
@@ -2,119 +2,157 @@
...
 * Migration upgrade -
 
-  1. Save snapshot of the the Cloudify Manager.
-  1. Install the Cloudify Manager on the host.
-  1. Restore snapshot of the Cloudify Manager to the host.
-  1. (Optional) Uninstall (teardown in Cloudify 4.2 and below) the Cloudify Manager from the host.
-  1. Migrate agents from the old Cloudify Manager.
+  1. [Backup a snapshot]({{< relref "installation/upgrading-manager.md#backup-a-snapshot" >}}) of the Cloudify Manager.
+  1. (For upgrade from 4.1.0 and below) If you have multiple Cloudify users, [backup user database]({{< relref "installation/upgrading-manager.md#backup-user-database" >}}).
+  1. (For upgrade from 4.0.0) [Backup the agent certificates]({{< relref "installation/upgrading-manager.md#backup-agent-certificates" >}}) from the Cloudify Manager.
```
geokala: This won't work for a migration upgrade.

bmansheim: Can you be more specific?

PR #587, opened 2018-01-31
Upload blueprint command

uriwygo:
Upload blueprint command should be without the -p flag. 
It can be uploaded either from an unzip file which you pass with the blueprint upload command: cfy blueprints upload -b BLUEPRINT_NAME  BLUEPRINT_YAML_FILE_PATH
Or by passing the command with -n flag that will indicate the blueprint's file in the zip file (can be zip from URL or from the VM)

### DIFF
--- a/content/manager/upload-blueprint.md
+++ b/content/manager/upload-blueprint.md
@@ -16,7 +16,7 @@ From the Cloudify command-line interface, you can upload your blueprint to Cloud
 
 The following is an example of `upload`:
 {{< gsHighlight  bash >}}
-$ cfy blueprints upload -b BLUEPRINT_ID -p BLUEPRINT_FILE_LOCATION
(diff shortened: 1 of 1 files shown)


content/manager/upload-blueprint.md
```diff
@@ -16,7 +16,7 @@ From the Cloudify command-line interface, you can upload your blueprint to Cloud
 
 The following is an example of `upload`:
 {{< gsHighlight  bash >}}
-$ cfy blueprints upload -b BLUEPRINT_ID -p BLUEPRINT_FILE_LOCATION
+$ cfy blueprints upload -b BLUEPRINT_ID -n BLUEPRINT_FILE_LOCATION URL_TO_BLUEPRINT_ARCHIVE
```
bmansheim: The explanation before this says "If you have downloaded cloudify-nodecellar-example from github..." so it looks like we just have to remove the -n from the command because it assumes that you already downloaded it to the manager.

What do you think?

uriwygo:
The -n flag works also if you give the URL zip, so you don't need to
download the zip file to the VM

PR #596, opened 2018-02-01
Deleted cluster host-ip flag

uriwygo:
Join cluster process could be without the host-ip flag. Moreover, when I tried to pass the flag in the join command it failed. Does it a supported command?


tehasdf:
@uriwygo that flag still exists in 4.3 and seems to work for me. How did it break?

PR #655, opened 2018-03-01
system resources page 4.3 update

edenpessach:


content/manager_webui/plugins-snapshots-page.md
```diff
@@ -10,20 +10,22 @@ The appearance of this page depends on your permissions. Only an `admin` can vie
 
 ## Plugins
 
-Plugins are tenant-specific because a blueprint on one tenant cannot access a plugin on a different tenant.  The Plugins table provides a list of all plugins that have been uploaded to the current tenant. For information about using default plugins or creating your own, [click here]({{< relref "plugins/overview.md" >}}).
+Plugins are tenant-specific by default unless uploaded as global-specific and therefore a blueprint on one tenant cannot access a plugin on a different tenant usually.  The Plugins table provides a list of all plugins that have been uploaded and available at the current tenant. For information about using default plugins or creating your own, [click here]({{< relref "plugins/overview.md" >}}).
```
bmansheim: How's this:
By default, [plugins]({{< relref "plugins/overview.md" >}}) are tenant-specific, meaning that a blueprint on one tenant cannot access a plugin on a different tenant. You can also [set a plugin as global]({{< "cli/plugins.md" >}}) when you upload it to the manager. The Plugins table lists the plugins are available to the current tenant.

content/manager_webui/plugins-snapshots-page.md
```diff
@@ -70,8 +72,8 @@ Depending on your requirements, the tenant may be prepared according to one of t
 
 ### Deleting a Snapshot
 
-1.  In the Snapshots widget, click the List icon on the far right of the snapshot entry that you want to delete.
-2. Click **Delete** to delete the snapshot from Cloudify Manager.
+1.  In the Snapshots widget, click the trashbin icon on the far right of the snapshot entry that you want to delete.
```
bmansheim: trashbin > trash bin
It would be better to use the icon instead of the name, such as:
... click [ ] for the snapshot entry that you want to delete.

content/manager_webui/plugins-snapshots-page.md
edenpessach: Sounds good, but without the link to the cli page.. this is the web interface docs, they can upload plugins in the ui as well

content/manager_webui/plugins-snapshots-page.md
edenpessach: I don't know how to show the icon here, if it's possible then I agree it's better

PR #658, opened 2018-03-01
Update secret store page with cli and web ui examples

inbala:


inbala:
There isn't a need, just thought it'll look nicer.
But maybe it should be consistent with other pages.

bmansheim:
Can you remove the <br> and put all of the lines in the paragraph together?

PR #661, opened 2018-03-04
Docs Check-Up

EarthmanT:
We need to have someone from R&D review this. I'm not sure if the information here is very old or very new. But it is surely very unclear and incomplete.

### DIFF
--- a/content/blueprints/import-resolver.md
+++ b/content/blueprints/import-resolver.md
@@ -9,54 +9,25 @@ weight: 1700
 
 An import resolver can be used to resolve imports during blueprint parsing.
 
-The Cloudify parser can accept different resolver implementations.
-It's possible to use Cloudify's default import resolver or to specify a new implementation that inherits from [AbstractImportResolver](https://github.com/cloudify-cosmo/cloudify-dsl-parser/blob/master/dsl_parser/import_resolver/abstract_import_resolver.py#L23)
+The Cloudify parser can accept different resolver implementations. It's possible to use Cloudify's default import resolver or to specify a new implementation that inherits from [AbstractImportResolver](https://github.com/cloudify-cosmo/cloudify-dsl-parser/blob/4.3/dsl_parser/import_resolver/abstract_import_resolver.py#L31)
 class and implements:
 
 * resolve(import_url) - returns the content of the resolved import URL.
-	* This method is called by the fetch_import method (implemented by the [AbstractImportResolver](https://github.com/cloudify-cosmo/cloudify-dsl-parser/blob/master/dsl_parser/import_resolver/abstract_import_resolver.py#L23)
+	* This method is called by the fetch_import method (implemented by the [AbstractImportResolver](https://github.com/cloudify-cosmo/cloudify-dsl-parser/blob/4.3/dsl_parser/import_resolver/abstract_import_resolver.py#L31)
 	class) during the imports parsing process, on each import URL that starts with `http`, `https` or `ftp`.
-
-# Declaration
-
-The resolver configuration is located in the manager blueprint under the cloudify configuration:
-{{< gsHighlight  yaml  >}}
-node_templates
-  ...
-  manager:
-    properties:
-      cloudify:
-        ...
-        import_resolver:
-          implementation: my_module.my_resolver:MyImportResolver
-          parameters:
-            param1: value1
-            param2: value2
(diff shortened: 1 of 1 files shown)


content/blueprints/import-resolver.md
```diff
@@ -9,54 +9,25 @@ weight: 1700
 
 An import resolver can be used to resolve imports during blueprint parsing.
 
-The Cloudify parser can accept different resolver implementations.
-It's possible to use Cloudify's default import resolver or to specify a new implementation that inherits from [AbstractImportResolver](https://github.com/cloudify-cosmo/cloudify-dsl-parser/blob/master/dsl_parser/import_resolver/abstract_import_resolver.py#L23)
+The Cloudify parser can accept different resolver implementations. It's possible to use Cloudify's default import resolver or to specify a new implementation that inherits from [AbstractImportResolver](https://github.com/cloudify-cosmo/cloudify-dsl-parser/blob/4.3/dsl_parser/import_resolver/abstract_import_resolver.py#L31)
```
bmansheim: Try not to use 's.
It's possible to use > You can use
use Cloudify's default import resolver > either:
- use the default import resolver
- use the Cloudify default import resolver

content/blueprints/import-resolver.md
```diff
@@ -9,54 +9,25 @@ weight: 1700
...
-[The default import resolver](https://github.com/cloudify-cosmo/cloudify-dsl-parser/blob/master/dsl_parser/import_resolver/default_import_resolver.py#L28)
+[The default import resolver](https://github.com/cloudify-cosmo/cloudify-dsl-parser/blob/4.3/dsl_parser/import_resolver/default_import_resolver.py#L28)
 is a default implementation of an import resolver.
-<br>This resolver is initialized with the ``rules`` parameter, which can be used later to replace import URL's prefix with another prefix
-and resolve the new URL (with the altered prefix).
 
-Each rule in the ``rules`` list is expected to be
-a dictionary with one (key, value) pair.
-Each rule represents a prefix and its replacement, which can be used to resolve the import URL.
+<br>This resolver is initialized with the ``rules`` parameter, which can be used later to replace import URL's prefix with another prefix and resolve the new URL (with the altered prefix).
```
bmansheim: Try not to use <br>.
This resolver starts with the ``rules`` parameter that you can also use to replace the import URL prefix with another prefix and resolve the new URL (with the altered prefix).

content/blueprints/import-resolver.md
```diff
@@ -9,54 +9,25 @@ weight: 1700
...
 
-The resolver will go over the rules in order to find a matching rule.
-For each matching rule, the resolver will replace the prefix denoted by the rule's key with the rule's value.
-The resolver will then attempt to resolve the new URL.
-If resolving is successful, the resolver will return the content of the resolved URL, otherwise it will try the next rule.
+Each rule in the ``rules`` list is expected to be a dictionary with one (key, value) pair. Each rule represents a prefix and its replacement, which can be used to resolve the import URL.
```
bmansheim: which can be used > that you can use

content/blueprints/import-resolver.md
```diff
@@ -9,54 +9,25 @@ weight: 1700
...
 
-If there aren't any rules, none of the rules matches or
-none of the prefix replacements could be resolved,
-the resolver will try to use the original URL.
+The resolver will go over the rules in order to find a matching rule. For each matching rule, the resolver will replace the prefix denoted by the rule's key with the rule's value. The resolver will then attempt to resolve the new URL. If resolving is successful, the resolver will return the content of the resolved URL, otherwise it will try the next rule.
```
bmansheim: Avoid "will"
The resolver looks at the rules to find a matching rule. For each matching rule, the resolver replaces the prefix denoted by the rule key with the rule value. The resolver attempts to resolve the new URL. If it successfully resolves, the resolver returns the content of the resolved URL. If it fails to resolve, it tries the next rule.

content/blueprints/import-resolver.md
```diff
@@ -9,54 +9,25 @@ weight: 1700
...
+
+If there aren't any rules, none of the rules matches or none of the prefix replacements could be resolved, the resolver will try to use the original URL.
```
bmansheim: If there are no rules, no matching rules or none of the prefix replacements resolve, the resolver uses the original URL.

PR #673, opened 2018-03-05
Add Insights widget docs

bmansheim:


bmansheim:
stage-docs.getcloudify.org/insights-widget/manager_webui/insights-widget/

qooban:
@talisela Insights widget is not a part of official Cloudify. Not sure, but maybe it's here: https://github.com/mistio/Cloudify-UI-Widget-boilerplate/tree/master/widgets/Insights
Do we want to add it to *cloudify-stage*?

bmansheim:
@qooban You're right. We'll put it somewhere else.

bmansheim:
Moved insights information to the manager section. Over the next couple of months we'll revamp the structure of the docs and find a better place to put this.

qooban:
@bmansheim For me it does not matter where that will be placed. We should only be sure that it is described as a widget which is not built-in in all versions.

PR #727, opened 2018-03-21
Update users for 4.3.

talisela:
These changes, once approved, should also be applied to 4.2 and 4.1 documentations.

### DIFF
--- a/content/cli/users.md
+++ b/content/cli/users.md
@@ -7,13 +7,17 @@ weight: 255
 ---
 
 The `cfy users` command is used to manage users and passwords on Cloudify Manager.<br>
-If you choose not to integrate Cloudify Manager with LDAP-based user management system, you must add each user individually and provide the users’ passwords. You can also create user groups and add users to them. The users and user groups can be assigned to one or more tenants, with different roles in each tenant. 
+If you choose not to integrate Cloudify Manager with LDAP-based user management system, you must add each user individually with a unique username and a password. You can also create user groups and add users to them. The users and user groups can be assigned to one or more tenants, with different roles in each tenant. 
 
 #### Requirements
 
 * To use the command you must have Cloudify `admin` credentials.<br>
-* Usernames and passwords must conform to the following requirements:  
-
+* Usernames must conform to the following requirements:
+  * Valid characters are alphanumeric, or `-`, `_`, or `.`.
+  * Value must begin with a letter
+  * Cannot be empty
+  
+* passwords must conform to the following requirements:  
   * Minimum number of characters - 5
   * Maximum number of characters - 255
   * Valid characters are alphanumeric, or `-`, `_`, or `.`.
@@ -46,7 +50,7 @@ Create a new user on Cloudify Manager.
 
 #### Optional flags
   
-* `-r, --security-role [sys_admin|default]` - A role to specify if the user is a sys-admin on the manager (in which case, the  role 'sys_admin' should be provided). If not, the user will be created as 'default', meaning you will needs to be explicitly assign it with tenants to access their resources. (default: default)
+* `-r, --security-role [sys_admin|default]` - A role that defines the user as a 'sys-admin' (admin user) or 'default' (non-admin user). A 'default' user must be explicitly assigned to tenants in order to perform actions and access resources. (default: default)
 
 &nbsp;
 #### Example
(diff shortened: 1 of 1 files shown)


content/cli/users.md
```diff
@@ -46,8 +46,7 @@ Create a new user on Cloudify Manager.
 
 #### Optional flags
   
-* `-r, --security-role [admin|user]` - A role to specifies the user's permissions
-                                  on the manager. (default: default)
+* `-r, --security-role [sys_admin|default]` - A role to specify if the user is a sys-admin on the manager (in which case, the  role 'sys_admin' should be provided). If not, the user will be created as 'default', meaning you will needs to be explicitly assign it with tenants to access their resources. (default: default)
```
bmansheim: A role that defines the user as a 'sys-admin' (admin user) or 'default' (non-admin user). A 'default' user must be explicitly assigned to tenants in order to perform actions and access resources.

content/cli/users.md
```diff
@@ -67,7 +66,7 @@ User `sue` created
 
 `cfy users set-password [OPTIONS] USERNAME`
 
-Set the password for a specific user. Use this command in a non-LDAP/AD setup.<br>
+Set the password for a specific user. You can use this command in a non-LDAP setup to change each of the users' passwords, and in LDAP mode to change only the password of the bootstrap-admin (the user created by default upon Manager's installation) .<br>
```
bmansheim: You can use this command in a non-LDAP setup to change each of the users' passwords, and in LDAP mode to change only the password of the bootstrap-admin (the user created by default upon Manager's installation) .<br>
->
You can use this command to change user passwords for non-LDAP users, and to change the password of the bootstrap-admin (the user created by default during Manager installation).

content/cli/users.md
```diff
@@ -93,30 +92,28 @@ New password set
 #### Usage 
 `cfy users set-role [OPTIONS] USERNAME`
 
-Set a role for a specific user. <br>
+Set the system-wide (security) role for a specific user. <br>
 
 `USERNAME` is the username of the user
 
-Users are created with the default system-wide `default` role. This command enables you to change a user's role to a Cloudify Manager administrator.
+the system-wide role of a user represents whether the user is a system admin on the manager, or not. To give the user sys-admin permissions, set this role to 'sys_admin'. Otherwise, the user should have the system-wide role 'default', which indicates that the user needs to be explicitly assigned to tenants to access their resources,
```
bmansheim: The system-wide role defines the user as a 'sys-admin' (admin user) or 'default' (non-admin user). To give the user sys-admin permissions, set this role to 'sys_admin'. Otherwise, the user has the system-wide role 'default'. A 'default' user must be explicitly assigned to tenants in order to perform actions and access resources.

content/cli/users.md
```diff
@@ -127,18 +124,23 @@ New role `admin` set
 #### Usage 
 ` cfy users delete [OPTIONS] USERNAME`
 
-Delete a user from Cloudify Manager, including from any groups to which they have been assigned.
+Delete a user from Cloudify Manager,
```
bmansheim: **Combine the lines like this:**
Delete a user from Cloudify Manager. You can delete a user only if the user is:

* Not assigned to any tenants
* Not a member of any user groups
* Not the creator of any Cloudify resources (Blueprint, Deployment, Plugin, Secret) on the Manager.

content/cli/users.md
```diff
@@ -231,7 +227,7 @@ Requested user info:
 #### Usage 
 `cfy users deactivate [OPTIONS] USERNAME`
 
-Deactivate a user. Suspends a user's access, without deleting their details.
+Deactivate a user. The user will not be able to log into Cloudify, but will not be deleted altogether. You can later on reactivate the user.
```
bmansheim: Deactivated users cannot login to Cloudify, but are in the list of users. To let the user login to Cloudify, reactivate the user.

content/cli/users.md
```diff
@@ -149,21 +151,19 @@ User removed
 #### Usage 
 `cfy users list`
 
-List all users defined in this Cloudify Manager.<br>
-By default, when you generate the list of users, only the number of linked resources are displayed. You can retrieve full details with the use of a `--get-data` flag.
+In a non-LDAP setup this command will list all users defined in this Cloudify Manager. In LDAP mode, it will list all users who have performed a log in into cloudify and were successfully authenticated against the LDAP-based system.<br>
```
bmansheim: Avoid future tense and "perform"
In non-LDAP mode, this command lists all of the users defined in this Cloudify Manager. In LDAP mode, this command lists all of the users who logged in to Cloudify and successfully authenticated with the LDAP system.

content/cli/users.md
talisela: I prefer separating it to "in LDAP mode" and "in non-LDAP mode", as a saying like "for non-LDAP users" might cause users to think you can have both LDAP and non-LDAP users at the same time.

PR #732, opened 2018-03-26
CYBL-198 HA: add the "finding the active manager" note

tehasdf:


bmansheim:
@tehasdf Please review my changes

bmansheim:
@tehasdf Reminder...

tehasdf:
@bmansheim thanks for the reminder, sorry for the delay!
looks good, but I've added one additional change - the original text, about the `True` value in the `master` column was only true until cloudify 4.2, but since 4.3, we've made it a `state` column that takes the values `leader`/`replica`/`offline`.
Ready to merge imo

PR #736, opened 2018-03-27
panel updates

shar1z:


bmansheim:
The files changed in the e231f61 commit are in the theme directory so it would take some effort to figure out what it is that you intended to change. Can you isolate your changes?

PR #751, opened 2018-04-10
put sudo in brackets

yeshess:


yeshess:
shouldn't have to run sudo for this and it could cause an issue with "docker load..."

bmansheim:
The syntax of the command has to be exact. I can write it without sudo and add a note that says: If you have a problem, you can try it with sudo.

yeshess:
No problem - add it as a note :)

PR #824, opened 2018-05-16
IP addresses should be configured prior to install

fifieldt:
Cloudify bootstrap does not configure the IP addresses on the private/public interfaces if they are not present, and if they are not configured the bootstrap will fail with an error about RabbitMQ connectivity. This is confusing for new users, reported multiple times eg  https://groups.google.com/forum/#!topic/cloudify-users/7gB8tIEv9E0 .

This patch adds a one line note to add configuration of IP addresses to pre-requisites, if they are not using a VM image-based install.


bmansheim:
This change is implemented this in a separate PR (https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/833).

Thanks for your input!

PR #853, opened 2018-06-06
Account lock docs

adigrabow:


bmansheim:
@adigrabow After we get the answer about the recommendation for cfy_manager configure I'll probably reorganize it a little.

bmansheim:
@adigrabow Please change the names of the fields to:

- failed_logins_before_account_lock
- account_lock_period

PR #908, opened 2018-06-28
Update default widgets descriptions for 4.4

talisela:


talisela:
Update default widgets for 4.4

talisela:
@qooban I was not aware of CYBL-214. However, I find it easier to look for a widget's description when they are listed alphabetically, as they do in the "all" category. What do you think?

PR #912, opened 2018-06-28
Updating cli's users commands docs

tyacbovi:


content/cli/maint_cli/users.md
```diff
@@ -46,7 +46,7 @@ Create a new user on Cloudify Manager.
 
 #### Required flags
   
-* `-p, --password TEXT` - Cloudify Manager password.
+* `-p, --password TEXT` - The user's password.
```
bmansheim: We try not to use possessives ('s). I recommend: The password for the user

PR #993, opened 2018-12-31
Restructure summaries in CLI

geokala:


content/cli/orch_cli/blueprints.md
```diff
@@ -310,6 +310,53 @@ Blueprints:
 ...
 {{< /highlight >}}
 
+### summary
+
+#### Usage 
+`cfy blueprints summary <target field> [optional sub-field] [OPTIONS]`
```
tyacbovi: I think that target in 'target field' is a little confusing, because you think target is something also you need to write

content/cli/orch_cli/blueprints.md
```diff
@@ -310,6 +310,53 @@ Blueprints:
...
+
+Gives a summary of blueprints based on the target field and an optional sub-field.
```
tyacbovi: I think it's unclear to define a term of summary with that term, I didnt understand what that means until I read the examples

geokala: Can you think of a clearer way to structure it? I couldn't while I was working through this, but I'll bounce it around in my head.

content/cli/orch_cli/blueprints.md
geokala: Hmm... better as target_field? Does 'optional sub-field' suffer from the same issue?

tyacbovi: Just field could be a fair choice, or summarized field maybe

content/cli/orch_cli/blueprints.md
tyacbovi: Maybe something like "gives a sum up of elements according to the summary by field"

PR #1197, opened 2019-11-28
vCloud plugin is only for vCloud Air, not for vCloud Director

twekkis:
vCloud plugin documentation updated to clearly state that plugin is only for Vmware vCloud Air.

vCloud Air seem not to be anymore (since 2017) in Vmware product portfolio. Plugin does not support
Vmware current vCloud Director, so documentation has to define that for the reader.


content/working_with/manager/create-deployment.md
```diff
@@ -127,7 +127,7 @@ The image is again the AMI image ID. The size is the instance_type, and the agen
 
   <!-- /gsInitContent -->
 
-  <!-- gsTabContent "vCloud " -->
+  <!-- gsTabContent "vCloud Air" -->
```
alexmolev:
```suggestion
  <!-- gsTabContent "vCloud" -->
```

PR #1216, opened 2020-01-02
Update node_id description

inbala:


content/cli/maint_cli/clusters.md
```diff
@@ -108,18 +108,25 @@ Cluster brokers
...
 
   Register a broker with the cluster.
 
   Note that this will not create the broker itself. The broker should have
   been created before running this command.
 
+#### Optional flags
+
+* `--port INTEGER RANGE` - A non-default network port to use for this component.
+* `-n, --networks TEXT` - Networks as a JSON string or as 'net1=ip1;net2=ip2'). This argument can be used multiple times.
```
ofercloudify: I think we are missing a bracket here.

inbala: Oh right it is the same in the cli

PR #1238, opened 2020-02-12
5.0.5 build

YoniItzhak:


alexmolev (changes requested):
The master is aligned with latest-build brach, overriding it with 505 will outdate the master branch

PR #1344, opened 2020-07-20
Cy 3150 upgrade to 510

AdarShaked:


content/ops_guides/backup_restore_guide.md
```diff
@@ -136,7 +137,36 @@ This step must take place before the cluster is created and before the snapshot
     1. {'status': 'Snapshot restore in progress...\nThis may take a while, depending on the snapshot size.'}
     1. {'status': 'No `restore_snapshot` workflow currently running.'}
 
+**If the restore is done as part of Upgrade to a newer Cloudify Manager version, consider perform also:**
```
glukhman, line 140: If the restore is done as part of Upgrade to a newer Cloudify Manager version, **the following should also be performed:**
(I don't think it's a matter for consideration. workflows which rely on agents won't work correctly)

content/ops_guides/backup_restore_guide.md
```diff
@@ -136,7 +137,36 @@ This step must take place before the cluster is created and before the snapshot
...
+
+1. Execute [install_new_agents workflow]({{< relref "working_with/workflows/built-in-workflows.md#the-install-new-agents-workflow" >}}) on the new Cloudify Manager so that all hosts agents are updated and connected to RabbitMQ on the new Cloudify Manager.
+
+1. Update plugins
```
glukhman, line 144: 2.
Ignore if it's auto-incremented by the docs engine

content/ops_guides/backup_restore_guide.md
```diff
@@ -136,7 +137,36 @@ This step must take place before the cluster is created and before the snapshot
...
 
+**If the restore is done as part of Upgrade to a newer Cloudify Manager version, consider perform also:**
+
+1. Execute [install_new_agents workflow]({{< relref "working_with/workflows/built-in-workflows.md#the-install-new-agents-workflow" >}}) on the new Cloudify Manager so that all hosts agents are updated and connected to RabbitMQ on the new Cloudify Manager.
+
+1. Update plugins 
+
+This is done in order to update the deployments to use new plugins(when upgrading to py2py3 plugins wagons).
+
+Firstly, upload new plugins, then execute:
```
glukhman, line 148: First*

content/ops_guides/backup_restore_guide.md
AdarShaked: well if no agents installed it will work correctly. That is the reason i wrote consider

PR #1424, opened 2020-09-22
CY-3438 Add warning re: keyfile path usage

glukhman:


alexmolev (approved):

EarthmanT:
@glukhman @alexmolev We should not even be supporting key files on the system. Our official supported method has been "only keys via secrets" since about 4.5.5.

alexmolev:
apparently there are still some customers that use it.... we are not supporting it, and even print a warning message.
I guess we will need to mark it as deprecated and probably remove it in the upcoming versiosn

PR #1528, opened 2020-10-27
Changing Cloudify Cluster Manager RPM installation link 5.1-build

YoniItzhak:


YoniItzhak:
Opened a new one instead

PR #1547, opened 2020-11-05
RD-293 Edit mode docs update

kubama:


qooban (approved):
Just one comment, a suggestion. Maybe it would be wiser not to include the header in the screenshot. It's painful to update all screenshots each release :-)

kubama:
Yes, of course - updated

PR #1554, opened 2020-11-13
RD-467 Update cluster management documentation

geokala:


geokala:
In progress, just ran into problems with cfy_manager commands with db proxy removed so slightly sidetracked on that now.

geokala:
I've got a final test run to do over these commands, but they're ready for review (for clarity, etc) now.

ofercloudify (approved):

content/ops_guides/ha_guides/cloudify_ha_day_two_ops.md
```diff
@@ -1,115 +1,210 @@
...
+```bash
+  cfy cluster status
+```
+The status should be healthy. Note that cluster status can take up to ~30 seconds to stabilise.
+
+### Set master
+
+If you wish to change the current DB master node, e.g. because the current master node is going to be undergoing maintenance operations, run the following command on a DB node:
+```bash
+  cfy_manager dbs set-master -a <intended new master's DB address as it appears in the node_ip field of the cfy_manager dbs list>
```
mabuaisha, line 202: @geokala I think its worth to mention that manager should be in maintenance mode before running this command, because I tried it and I got the following error message
```
2020-12-13 10:21:48,354 - [CONFIG] - INFO - Loading configuration from /etc/cloudify/config.yaml
2020-12-13 10:21:49,610 - [POSTGRESQL_SERVER] - INFO - 172.22.0.11 is async replica. Only sync replicas can become master.
2020-12-13 10:21:49,610 - [POSTGRESQL_SERVER] - INFO - Attempting to promote 172.22.0.11 to sync replica.
2020-12-13 10:22:02,207 - [MAIN] - ERROR - DBManagementError: 172.22.0.11 did not manage to become synchronous replica. Before you retry, please ensure the DB cluster is healthy and the managers are in maintenance mode.
DBManagementError: 172.22.0.11 did not manage to become synchronous replica. Before you retry, please ensure the DB cluster is healthy and the managers are in maintenance mode.
```

mabuaisha: This happened when trying to make async replica as master node but it failed, so what I did after that is to run change the ip address for sync replica and the leader get changed as expected

geokala: I thought that should retry- poke me on Monday and I'll check the code.

As for mentioning that the managers should be in maintenance mode, that's in a note at the top of the document.

ofercloudify (approved):

PR #1561, opened 2020-11-16
RD-235 Added overrides.json usage description

kubama:


content/working_with/console/customization/overrides.md
```diff
@@ -0,0 +1,19 @@
...
+draft: false
+weight: 180
+---
+
+You can override a number of texts and labels used in {{< param cfy_console_name >}} by supplying a file containing replacements for default texts or labels.
+
+To do that:
+
+1. Create `/opt/cloudify-stage/dist/userData/overrides.json` file.
+1. Go to https://github.com/cloudify-cosmo/cloudify-stage/blob/master/app/translations/en.json, choose configurations parts to modify (you can omit all unmodified parts to use default values) and save them to the created file.
```
qooban, line 14: As labels can change over time, to be more precise we can also write about an option to switch to specific release (by selecting a branch `${version}-build`).

content/working_with/console/customization/overrides.md
```diff
@@ -0,0 +1,19 @@
...
+1. In the browser reload {{< param cfy_console_name >}} to see the changes.
+
+{{% note %}}
+When setting up a {{< param cfy_manager_name >}} Cluster, these changes should be applied to every {{< param cfy_manager_name >}} in the cluster.
+{{% /note %}}
```
qooban, line 19: We can mention that, but as we don't do that in other places, maybe it's not necessary.
The whole `userData` folder is being replicated by `syncthing` tool (as far as I know).

(I mentioned recently about moving all files to DB, but if that happens, then we'll have to update docs anyway)

qooban (commented):
I think that we need to have few high-level sentences in the main Customization page as well (https://github.com/cloudify-cosmo/docs.getcloudify.org/edit/RD-235-i18n/content/working_with/console/customization/_index.md). 

It seems to me, that Layout level / Customizing Layout sections can be extended with few words about ability to change text in the application and a link to the page you created.
It would be good to mention there that text labels as well as external links in Cloudify Console are customizable (we can give some most important examples: header, product title, links to external data stores, etc.).

content/working_with/console/customization/overrides.md
kubama: > We can mention that, but as we don't do that in other places, maybe it's not necessary.
> The whole `userData` folder is being replicated by `syncthing` tool (as far as I know).

Actually the same note is present in the preceding 'User Configuration' section, which is also related to placing file in `userData`.

qooban: Oops ;-) My bad. OK, then leave it.

kubama:
Comments addressed - please review

qooban (approved):

content/working_with/console/customization/overrides.md
```diff
@@ -0,0 +1,19 @@
...
+draft: false
+weight: 180
+---
+
+You can override a number of texts and labels used in {{< param cfy_console_name >}} by supplying a file containing replacements for default texts or labels.
+
+To do that:
+
+1. Create `/opt/cloudify-stage/dist/userData/overrides.json` file.
+1. Go to https://github.com/cloudify-cosmo/cloudify-stage/blob/<version>-build/app/translations/en.json (replace `<version>` with your product version, for example `5.1.0`), choose configurations parts to modify (you can omit all unmodified parts to use default values) and save them to the created file.
```
ofercloudify, line 14: I dont think it is a good practice to let the user set the version in the URL, and then get the file from git.

Do we have such a file on the manager? can the user copy it from the folder where it resides?

ofercloudify, line 14: We should probably add a sample of a few lines from that file.

kubama: We don't have this file on manager, so the user can't just copy it.

Please note that although the link points to github it is a regular HTTP link that can be clicked and opened in a browser.
My initial commit was to link directly to `master` branch without the need for link editing. Should I revert to that? Or maybe just use `5.1.1-build` branch?

kubama: Done

qooban: @ofercloudify That's important question as well for me doing changes in Composer right now (RD-233). Translation file is not exposed in the production code there as well, it is bundled.

If we want to have it on the manager, let us know. 
We would need to add those files to the Stage and Composer package.

ofercloudify: OK, if this is the case we will stay with github.
I suppose the right approach to the version should be by adding a parameter to the config.toml for the version number and embed it in the link. The downside would be updating it with every release but that is on me.

ofercloudify (approved):

PR #1614, opened 2020-12-10
3rd party doc

ofercloudify:


qooban (approved):
Approving it, but I assume that probably we'll need an update with packages from section #2 (missing deps), right?

PR #1702, opened 2021-02-01
RD-981 Improve snapshots restore docs

YoniItzhak:


content/ops_guides/backup_restore_guide.md
```diff
@@ -124,13 +130,17 @@ Snapshots of the HA {{< param cfy_manager_name >}} cluster should be taken at re
 
     There are two possible responses:
     1. {'status': 'Snapshot restore in progress...\nThis may take a while, depending on the snapshot size.'}
-    1. {'status': 'No `restore_snapshot` workflow currently running.'}
+    1. {'status': 'No `restore_snapshot` workflow currently running.'}
```
ofercloudify, line 133: I dont understand this second status. If you read it as written it sounds like there is no restore snapshot process running, which is a weird error.
If a comma is missing and it is "No, a 'restore snapshot' workflow is currently running, please wait for it to complete." then it makes more sense.
not sure what the  writer's intention was...

YoniItzhak: Good comment. The meaning is that a `restore snapshot` workflow is not currently running, so either a snapshot restore happened or not, it's not currently running.

PR #1748, opened 2021-03-21
RD-1706-Fabric-plugin-non-recoverable-errors

ahmadiesa-abu:
added the new input description


EarthmanT (approved):

content/working_with/official_plugins/Configuration/fabric.md
```diff
@@ -435,6 +435,35 @@ node_templates:
 {{< /highlight >}}
 
 
+# Exception Handling (2.X version only)
+
+Fabric might generate exception if there is a problem with the execution. By default all of that exceptions are considered recoverable but you can control that with a provided list of error codes to change that to be non-recoverable. The `non_recoverable_error_exit_codes` input is a list of `exit_codes`.
```
ofercloudify, line 440:
```suggestion
Fabric might generate an exception if there is a problem with the execution. By default, all of the exceptions are considered recoverable but you can control that with a provided list of error codes to change that to be non-recoverable. The `non_recoverable_error_exit_codes` input is a list of `exit_codes`.
```

ofercloudify (changes requested):

ofercloudify (approved):

PR #1830, opened 2021-05-13
RD-2318 - Improve Filters widget description

qooban:
Added operator mapping UI/API/CLI to Filters widget description (https://docs.cloudify.co/staging/RD-2318/working_with/console/widgets/filters):

image

### DIFF
--- a/content/working_with/console/widgets/filters.md
+++ b/content/working_with/console/widgets/filters.md
@@ -40,8 +40,27 @@ Add, edit and clone operation modals share a common component for defining filte
 The component presents a list of rows, each representing a single filter rule. Each row contains three inputs:
 
 * Rule type selection dropdown - selecting the context of the rule which can be based on labels or supported deployment attributes such as blueprint ID, site name, or creator.
-* Rule operator dropdown. The set of available operators to choose from depends on the selected rule type.
+* Rule operator dropdown. The set of available operators to choose from depends on the selected rule type. See Table 1. and Table 2. below for details.
 * Value input (for attribute rules) or key/value input(s) (for label rules). 
+
+| Operator label in the UI       | Operator name in the API/CLI        |
+|:------------------------------:|:-----------------------------------:|
+| is one of                      | `any_of`                            |
+| is not one of                  | `not_any_of`                        |
+| is not one of (or no such key) | `is_not`                            |
+| key is not                     | `is_null`                           |
+| key is                         | `is_not_null`                       |
+**Table 1**. Labels operators
+
+| Operator label in the UI  | Operator name in the API/CLI        |
+|:-------------------------:|:-----------------------------------:|
+|   is one of               | `any_of`                            |
+|   is not one of           | `not_any_of`                        |
+|   contains                | `contains`                          |
+|   does not contain        | `not_contains`                      |
+|   starts with             | `starts_with`                       |
+|   ends with               | `ends_with`                         |
+**Table 2**. Attributes (Blueprint, Site name, Creator) operators mapping
 
 At any given time it is possible to append a new rule to the list of already defined rules (by clicking `Add new rule` button) or to remove any rule by clicking the trash icon in the corresponding rule row (unless there is only single rule defined).
 
--- a/content/working_with/console/widgets/filters.md
+++ b/content/working_with/console/widgets/filters.md
@@ -40,31 +40,107 @@ Add, edit and clone operation modals share a common component for defining filte
 The component presents a list of rows, each representing a single filter rule. Each row contains three inputs:
 
 * Rule type selection dropdown - selecting the context of the rule which can be based on labels or supported deployment attributes such as blueprint ID, site name, or creator.
-* Rule operator dropdown. The set of available operators to choose from depends on the selected rule type. See Table 1. and Table 2. below for details.
+* Rule operator dropdown. The set of available operators to choose from depends on the selected rule type. See **Table 1.** and **Table 2.** below for details.
 * Value input (for attribute rules) or key/value input(s) (for label rules). 
 
-| Operator label in the UI       | Operator name in the API/CLI        |
-|:------------------------------:|:-----------------------------------:|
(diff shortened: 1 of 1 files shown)


qooban:
@ofercloudify Please take a look. I would also appreciate taking a look at Filters and Labels widgets description in general and if you see that we should describe something in more detail I can do that next week.

content/working_with/console/widgets/filters.md
```diff
@@ -40,9 +40,28 @@ Add, edit and clone operation modals share a common component for defining filte
 The component presents a list of rows, each representing a single filter rule. Each row contains three inputs:
 
 * Rule type selection dropdown - selecting the context of the rule which can be based on labels or supported deployment attributes such as blueprint ID, site name, or creator.
-* Rule operator dropdown. The set of available operators to choose from depends on the selected rule type.
+* Rule operator dropdown. The set of available operators to choose from depends on the selected rule type. See Table 1. and Table 2. below for details.
 * Value input (for attribute rules) or key/value input(s) (for label rules). 
 
+| Operator label in the UI       | Operator name in the API/CLI        |
```
kubama, line 46: Just wondering.  We are using 'label' word as a term for cfy label entity, for example in L44 ('label rules'). But here we are using it as a term for string occurring in the UI. Isn't it confusing? Wouldn't it be better to name this table column 'Operator name in the UI'?

ofercloudify: +1 
I had the same comment :)

ofercloudify, line 46: I think we should also include the logic. Here is my suggestion for the tables:

| Operator name in the UI | Operator name in the API/CLI | Applied logic |
|:------------------------------:|:-----------------------------------:|:--------:|
| is one of                      | `any_of`                            | The label key matches the specified key and the label value matches one of the specified values. |
| is not one of                  | `not_any_of`                        | The label key matches the specified key and the label value does not match any of the specified values. |
| is not one of (or no such key) | `is_not`                            | No label key matches the specified key, or the label key matches the specified key and the label value does not match any of the specified values. |
| key is not                     | `is_null`                           | No label key matches the specified key. |
| key is                         | `is_not_null`                       | The label key matches the specified key. |
**Table 1**. Labels operators

| Operator name in the UI  | Operator name in the API/CLI        | Applied logic |
|:-------------------------:|:-----------------------------------:|:------------:|
|   is one of               | `any_of`                            | The deployment property matches one of the specified values. | 
|   is not one of           | `not_any_of`                        | The deployment property does not match any of the specified values. |
|   contains                | `contains`                          | The deployment property contains the specified value. |
|   does not contain        | `not_contains`                      | The deployment property does not contain the specified value. |
|   starts with             | `starts_with`                       | The deployment property starts with the specified value. |
|   ends with               | `ends_with`                         | The deployment property ends with the specified value. |
**Table 2**. Attributes (Blueprint, Site name, Creator) operators mapping

qooban: Make sense. I have refactored it even more 🙂 Hope it's more user-friendly now.

qooban: Thanks for the detailed description. I have applied it with some changes (`deployment property` -> `deployment attribute`).

qooban:
@ofercloudify @kubama I have applied your comments and went even further:

1. As I found that actually the operators naming in the CLI is not the same as in the API, I have refactored the table to also show that difference. I think right now we have quite accurate description. I'm wondering that maybe it should be placed somewhere else (but for now we can keep it in Filters widget description page).

2. Also, I changed the table format from simple markdown table to HTML table. That allowed me to add better styling - make columns width to differ and not be equal. It doesn't make sense to have equal widths as the description needs to have more space and API/CLI operators are short.

kubama:
Great job!
I have two comments:
- The 'APPLIED LOGIC' column header is in upper case, is this due to the default styling? Can it be changed?
- I think operators logic description should be placed in the core functionality section in the first place. Is it present there? If yes maybe it could be just referenced here not to duplicate the content? If not maybe it's a good idea to move it there?

qooban:
> * The 'APPLIED LOGIC' column header is in upper case, is this due to the default styling? Can it be changed?

It's default styling from Hugo framework I suspect. See other tables in the documentation. 
It can be changed of course, but IMHO if other tables use such styling, I think we can keep it for consistency.

> * I think operators logic description should be placed in the core functionality section in the first place. Is it present there? If yes maybe it could be just referenced here not to duplicate the content? If not maybe it's a good idea to move it there?

Yes, I was also thinking about that. This is what I meant by writing "I'm wondering that maybe it should be placed somewhere else".
In https://docs.cloudify.co/staging/RD-2318/cli/orch_cli/filter-rules/ we have some description about the operators. That page is referenced from Filters description page, so maybe that would be enough and we can drop "Applied logic" column? 

I would leave that to Ofer to decide.

ofercloudify:
Hey, really great work on that. thanks!
Adding @YoniItzhak as I believe there is value in copying this to the core help page of filters.

@qooban @kubama , I rather not link to it but rather have it in the widget help because this way people who are not sure and click the widget (?) will be able to view it directly.

ofercloudify (approved):

YoniItzhak:
@ofercloudify, will do.

YoniItzhak:
@qooban, the documentation only refers to deployments' filters and labels. Are we not referring to blueprints' filters and labels intentionally?

qooban:
@YoniItzhak I used suggestion from Ofer and didn't think about different filters. 
On the other side - I actually was not aware of blueprints' filters. How blueprints' filters work? Where can we use it?

ofercloudify:
The blueprint labels are much more limited in terms of usage. right now the plan was to only include them in the local blueprints page as a widget level filter but it seems like we missed it.
I am adding it to 6.1. therefore no docs for it yet.

YoniItzhak:
@ofercloudify, thank you. I'll consider it when copying @qooban's table to my documentation.

PR #1879, opened 2021-06-15
RD-1075 Updated widget APIs documentation

kubama:


qooban (approved):
Changes seems good. Just one reminder - we should clearly notify Alex/Ofer about coming braking change in the Widgets API and IMHO provide a way for customers to migrate.

ofercloudify (approved):

PR #1907, opened 2021-07-29
RD-2775 update after jQuery removement.

stepek:
Remove jquery, $ api and postRender method references.

### DIFF
--- a/content/developer/writing_widgets/widget-apis.md
+++ b/content/developer/writing_widgets/widget-apis.md
@@ -302,7 +302,7 @@ Redirects user to parent page (used when you are in drill-down page).
 
 Shows/hides a loading spinner in widget header.
 
-**Not allowed in `render()` and `postRender()`** methods as it changes store state leading to `render()` and `postRender()` re-run.
+**Not allowed in `render()`** method as it changes store state leading to `render()` re-run.
 
 
 ### refresh()
@@ -335,7 +335,7 @@ To see what functions are available in latest version see source code: [main fil
 
 ## External Libraries
 
-The external libraries available to a widget are: `React`, `PropTypes`, `moment`, `jQuery`, `Lodash`.
+The external libraries available to a widget are: `React`, `PropTypes`, `moment`, `Lodash`.
 
 You can assume that the following names are globally available as they are attached to the browser window object.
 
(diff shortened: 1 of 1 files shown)


content/developer/writing_widgets/widget-apis.md
```diff
@@ -302,7 +302,7 @@ Redirects user to parent page (used when you are in drill-down page).
 
 Shows/hides a loading spinner in widget header.
 
-**Not allowed in `render()` and `postRender()`** methods as it changes store state leading to `render()` and `postRender()` re-run.
+**Not allowed in `render()`** method as it changes store state leading to `render()` re-run.
```
Gelio, line 305: In case the intent is to remove all information about `postRender`, then I found some more occurrences:

```sh
13:15 $ rg -i postrender content
content/developer/writing_widgets/widget-definition.md
48:`fetchUrl`             | string/object | No | -            | If `fetchUrl` exists, the data from the URL is fetched by the application and passed to the render and postRender methods. To fetch multiple URLs, you must pass an object where the key is a name you select for this data, and the value is the URL. It is important to note that the render is called once before the data is fetched (to enable information about loading or partial data can be displayed) and once after the data is fetched.
400:### postRender(container, widget, data, toolbox)
404:`postRender` is called immediately after the widget has been made visible in the {{< param cfy_console_name >}}.
```

Is `postRender` deprecated? I don't think we are removing it :thinking: It's still there in the types

image

Gelio (commented):
Generally LGTM, but I'm slightly confused about `postRender` changes, thus, no approve yet :smile:

content/developer/writing_widgets/widget-apis.md
qooban: `postRender` should not be removed from the documentation. As @Gelio pointed, it is still in use.

@stepek , you probably was misled by the description I provided in the RD-2775 (as it was created when planned to remove it and I didn't track it afterwards).

qooban (commented):
I see that CI does not like the way how you name branches 🙂 So you need to update command in Jenkinsfile or recreate the change on different branch.

Minor thing - typo to fix in the title - there's no such word as "removement" (https://wikidiff.com/removement/removal) 🙂

stepek:
https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1909

PR #1942, opened 2021-09-01
RD-3004 Document get_attributes_list intrinsic

geokala:


content/developer/blueprints/spec-intrinsic-functions.md
```diff
@@ -616,6 +616,50 @@ This limitation has significant implications when using `get_attribute` in node/
...
+
+`get_attributes_list` is used to reference runtime-properties of all node instances of a target node from within a blueprint.
+
+## Usage and Examples
+### get_attributes_list general behaviour
+`get_attributes_list` behaviour is the same as `get_attribute` in most respects- e.g. with regards to evaluation always resulting
+in the current values and the behaviour with nested structures.
+
+The behaviour differs in that `get_attributes_list` will always return a list of the target attribute for all node instances belon
+ging to the target node. Therefore, it is usable with nodes with multiple instances without needing to be part of a scaling group.
```
tehasdf, line 629:
```suggestion
The behaviour differs in that `get_attributes_list` will always return a list of the target attribute for all node instances belonging
to the target node. Therefore, it is usable with nodes with multiple instances without needing to be part of a scaling group.
```

content/developer/blueprints/spec-intrinsic-functions.md
```diff
@@ -616,6 +616,50 @@ This limitation has significant implications when using `get_attribute` in node/
...
+
+### Notes, Restrictions and Limitations
+These are identical to the `get_attribute` function with the exception that there is no limitation regarding a target node's insta
+nce count.
```
tehasdf, line 633:
```suggestion
These are identical to the `get_attribute` function with the exception that there is no limitation regarding a target node's instance count.
```

content/developer/blueprints/spec-intrinsic-functions.md
```diff
@@ -616,6 +616,50 @@ This limitation has significant implications when using `get_attribute` in node/
...
+
+If any node instances are missing the requested attribute, there will be no value added to the list (null or otherwise) for that n
+ode instance (unless the node itself has the property per the lookup rules noted under `get_attribute`). If all node instances are missing the requested attribute then an empty list will be returned.
```
tehasdf, line 636:
```suggestion
If any node instances are missing the requested attribute, there will be no value added to the list (null or otherwise) for that node instance
(unless the node itself has the property per the lookup rules noted under `get_attribute`). If all node instances are missing the requested attribute then an empty list will be returned.
```

tehasdf: actually, now that I think of it, should it not return nulls instead? RD-3004 says:
`The number of items in the list is always matching the number of node-instances`
which suggests a missing-value thing

content/developer/blueprints/spec-intrinsic-functions.md
```diff
@@ -616,6 +616,50 @@ This limitation has significant implications when using `get_attribute` in node/
...
+node_templates:
+  web_server
+    type: cloudify.nodes.WebServer
+    instances:
+      deploy: 3
+
+outputs:
+  web_server_ips:
+    description: Web server IPs
+    value: { get_attribute: [web_server, ip] }
```
tehasdf, line 653:
```suggestion
    value: { get_attributes_list: [web_server, ip] }
```

content/developer/blueprints/spec-intrinsic-functions.md
geokala: Ah yes, you're right- I misread that part of the jira. Will update.

content/developer/blueprints/spec-intrinsic-functions.md
geokala: Good catch

content/developer/blueprints/spec-intrinsic-functions.md
geokala: Fixed, copy-pasta error.

content/developer/blueprints/spec-intrinsic-functions.md
geokala: Updated docs, code update coming soon™.

MichaelShn (approved):

PR #1980, opened 2021-10-06
RD-3313 emphasize community/premium version

warolv:
RD-3313 emphasize community/premium version


natishalom:
can you move it to be under the community version
the title should be: Cloudiy as Kubernetes Service

PR #2046, opened 2021-12-02
RD-3477 Removed gramatically incorrect prepending LDAP with 'the'

NorbertAmartus:
This change is related to the discussion which took place [here](https://github.com/cloudify-cosmo/cloudify-stage/pull/1682#discussion_r759250926)

The result of the conversation was that the `LDAP` is incorrectly prepended with the `the` word


NorbertAmartus:
I've created the PR once more, because as I can recall, it is impossible to change the branch name without closing the existing PR :/

kubama (approved):

PR #2072, opened 2022-01-11
Rd 3755 document agent install provided

glukhman:


content/install_maintain/agents/installation.md
```diff
@@ -130,24 +130,30 @@ every agent.
    node instance, or form the deployment logs:    
    ![Script in logs]( /images/manager/agent_installation/logs_script.webp )
 1. Login to the agent's VM.
+1. Prepare the agent directory
+   ```shell
+   sudo mkdir -p /opt/cloudify-agent-6.3.0-ga/
+   sudo chown $(whoami) /opt/cloudify-agent-6.3.0-ga/
```
geokala, line 136: Why set it to be owned by the user? Shouldn't it be root?

content/install_maintain/agents/installation.md
```diff
@@ -130,24 +130,30 @@ every agent.
...
+   mkdir -p /opt/cloudify-agent-6.3.0-ga/container_nwxqiu/cloudify/ssl
```
geokala, line 137: Anywhere you're using the agent's identified (container_nwxqiu), you should probably be using a stirng that is noted at the beginning?

content/install_maintain/agents/installation.md
```diff
@@ -130,24 +130,30 @@ every agent.
...
    ("container_nwxqiu" in the above example):
    ```shell
-   tar xzfv centos-core-agent.tar.gz --strip=1 -C ~centos/container_nwxqiu
+   tar xzfv centos-core-agent.tar.gz --strip=1 -C /opt/cloudify-agent-6.3.0-ga/container_nwxqiu
    ```
-1. Download the configuration script using the like from step 4:
+1. Download the configuration script using the link from step 4:
    ```shell
-   wget --no-check-certificate --user admin --password admin https://10.239.0.207:53333/resources/cloudify_agent/3631bbc3-fc37-4a9f-9db3-81adb2d8e182.sh
+   wget --no-check-certificate --user admin --password admin https://10.239.x.y:53333/resources/cloudify_agent/3631bbc3-fc37-4a9f-9db3-81adb2d8e182.sh
```
geokala, line 146: Just use RFC5737 compliant IPs- e.g. https://192.0.2.123:53333/etc/etc

geokala (approved):

content/install_maintain/agents/installation.md
glukhman: otherwise the script won't work, even with sudo

geokala: That doesn't sound right- I thought we set that dir to be owned as root when we installed agents. How does the script fail?

If we don't do that then presumably trying to reinstall an agent as a different user will result in fire and sadness.

geokala: After further discussion, we're going to leave this part of the instructions alone for the moment as it should only have impact if installing multiple agents on the same machine using this approach.

geokala (approved):

PR #2086, opened 2022-02-01
RD-3191 Document that the deployment schedules UI part is still to be…

glukhman:
… developed


tehasdf (approved):
Please also figure out a way to write down that this needs to be removed or updated after UI actually implements this feature
@qooban @alexmolev

qooban (approved):
> Please also figure out a way to write down that this needs to be removed or updated after UI actually implements this feature

The most reasonable way IMHO is to add it to relevant JIRA ticket. The best would be to have epic for Deployment Schedules in the UI. Then, IMHO adding a short note in ticket for docs update would be sufficient. But as for now I don't see any issues for Deployment Schedules in the UI. RD-1052 was for backend part, right? @alexmolev , @ilanyaniv - what is the plan about this feature?

PR #2091, opened 2022-02-03
RD-1923 Updated widget backend development doc after adding support for TS

kubama:
The doc contains reference to https://github.com/cloudify-cosmo/Cloudify-UI-Widget-boilerplate/tree/master/widgets/backendWidget which is pretty outdated. There are tickets like RD-229 and RD-231 which should address this, so I assume no further action is required ATM.

### DIFF
--- a/content/developer/writing_widgets/widget-backend.md
+++ b/content/developer/writing_widgets/widget-backend.md
@@ -21,15 +21,15 @@ Example of working widget with backend can be found [here](https://github.com/cl
 
 ## Defining Endpoints
 
-To create endpoint per widget you need to create `backend.js` file with at least one endpoint definition. That file must be placed in widget main folder similarly to `widget.js` file.
+To create endpoint per widget you need to create `backend.ts` file with at least one endpoint definition. That file must be placed in widget main folder similarly to `widget.js` file.
 
 
-### `backend.js` file structure
+### `backend.ts` file structure
 
-Example of `backend.js` file is presented below:
+Example of `backend.ts` file is presented below:
 
-```javascript
-module.exports = function(r) {
+```typescript
+export default function(r) {
 
     r.register('manager', 'GET', (req, res, next, helper) => {
         let _ = require('lodash');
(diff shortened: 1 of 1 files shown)


jawamartus (commented):
General question.
IIUC currently we accept widgets written in js and ts files. Is that true?

Do we still accept widgets written with `backend.js` files after last changes?

content/developer/writing_widgets/widget-backend.md
```diff
@@ -21,15 +21,15 @@ Example of working widget with backend can be found [here](https://github.com/cl
 
 ## Defining Endpoints
 
-To create endpoint per widget you need to create `backend.js` file with at least one endpoint definition. That file must be placed in widget main folder similarly to `widget.js` file.
+To create endpoint per widget you need to create `backend.ts` file with at least one endpoint definition. That file must be placed in widget main folder similarly to `widget.js` file.
```
jawamartus, line 24: Do you think that it is reasonable to change also `widget.js` to `widget.ts`?

kubama: No, as there is no change in this area.

jawamartus:
> As noted above, this change affects only widget backend. The file has to be named `backend.ts`, but since TS is a superset of JS it can still contain pure JS.

I see some risk here. The change is not back-compatible. What about people who have own custom widgets written in backend.js which are not present in our repository? It is possible to use cloudify-ui/console this way? Do cloudify have customer who have custome widgets not present in our repository?

jawamartus (approved):
LGFM, however please answer the question I asked above related to back-compatibility.

kubama:
> I see some risk here. The change is not backward-compatibility. What about people who have own custom widgets written in backend.js which are not present in our repository? It is possible to use cloudify-ui/console this way? Do cloudify have customer who have custome widgets not present in our repository?

This might be a valid point, however I'm not sure if there is a single customer using this feature. @qooban Are you aware of any? If we know that there is (or we are not sure) then we could add backward compatibility relatively easy. This will require further changes in stage, so I think it could be handled as part of separate ticket. In this case I would merge this PR and create another ticket, if necessary. Lets wait for @qooban's opinion though.

qooban:
@jawamartus @kubama Consulted with @alexmolev and the decision is to keep backward compatibility as adding it should be relatively easy.

In the coming weeks I'm going to investigate using manager's usage collector mechanism to get more data about widgets (and custom widgets) usage by customers. That hopefully will make it easier for us to make decisions.

PR #2095, opened 2022-02-08
RD-4194: Updated description of the secrets selection

NorbertAmartus:
Documentation update is related to the changes made in [this PR](https://github.com/cloudify-cosmo/cloudify-stage/pull/1804).
Documentatiton modification should be synchronised also as a part of [this PR](https://github.com/cloudify-cosmo/cloudify-stage/pull/1804) 😉


content/working_with/console/widgets/blueprints.md
```diff
@@ -75,7 +75,7 @@ The default visibility is "Tenant", and according to the logged-in user's permis
       - `Value / Secret key / Input name` - value of the variable (related to the selected `Source`).<br />
          Selecting `Static` as a `Source` enables to type a value. <br />
          Selecting `Input` as a `Source` enables to type a value. <br />
-         Selecting `Secret` enables to select a value from one of the stored secrets.
+         Selecting `Secret` enables to select a value from one of the stored secrets or to enter not existing one.
```
kubama, line 78: This sounds quite surprising to me 🙂.
The idea behind a 'secret' is a reusable entity that can be used by referencing it when it's defined. So, what will happen if the user selects 'not existing one'? And what 'one' refers to - a secret or a secret value?

NorbertAmartus: It refers to the secret name, and for now it's just adding a reference to the secret which is not existing
Michael showed me, that currently there was an option to do just that, by typing any value as a `source === static` and switch the `source` to the `secret`. Value would be preserved, and therefore we had an ability to use the secret which was not yet created 😁 (definitely a bug on our side)
And as I understand, he wanted to make it more "legal" to use that method, without those little tricks

So having that said, it's working the same way as previously (as it was an option to do so 😁), but without the stored secret.
So the blueprint will be referencing to the secret, which may be created in the future

It's a quite surprising behaviour for me too, as previously we were only allowed to access the secrets across the system, which were already defined (as you said)
But it seems like a business thought about it is being changed ... in a surprising way 😁

kubama: Thanks for a clear explanation. As the idea is a little bit confusing I would like to propose a little bit more accurate description:
```suggestion
         Selecting `Secret` enables to choose one of the stored secrets as the variable value source or to manually enter secret key, which may not be defined in the secret store yet.
```

NorbertAmartus: No problem and sure, it definitelly looks more precisely 🎉

kubama (approved):

PR #2183, opened 2022-04-27
RD-4759 Describe `item_type` property of inputs and workflow parameters

mateumann:


tehasdf (approved):

content/developer/blueprints/spec-inputs.md
```diff
@@ -36,6 +36,7 @@ inputs:
...
 | type          | no       | string        | The required data type of the input. Not specifying a data type means the type can be anything, including a list, an array or a dictionary. Valid types: `string`, `integer`, `float`, `boolean`, `list`, `dict`, `regex`, `textarea`, `blueprint_id`, `deployment_id`, `secret_key`, `capability_value`, `scaling_group`, `node_id`, `node_type`, `node_instance` or a [custom data type]({{< relref "developer/blueprints/spec-data-types.md" >}}). |
+| item_type     | no       | string        | Definition of items' type, only valid for `list` type, if none is provided the items' type can be anything.                                                                                                                                                                                                                                                                                                                                           |
```
geokala, line 39: item's if it's the type of one item

geokala: And if it's more than one item then it should be "items' types"

mateumann: I thought of that and since list can have (potentially) more items, I left it as _items'_

geokala: Hmm... or maybe it shouldn't be- yes, this might be correct, but is it?

geokala: Fair enough, marking this as resolved then.

PR #2217, opened 2022-05-10
RD-4616 workflows: note about availability_rules

tehasdf:
Currently only describe the simplest rule, `available`. Further rules
will be described when they're implemented


tehasdf:
Staging https://docs.cloudify.co/staging/workflow-availability-1/developer/blueprints/spec-workflows/

glukhman (approved):

PR #2218, opened 2022-05-12
RD-4618 cfy workflows list: note on availability

tehasdf:
- there's now also an `--all`
- the created_at column is gone, but availability_rules now exists


tehasdf:
staging https://docs.cloudify.co/staging/workflow-availability-2/cli/orch_cli/workflows/

mateumann (approved):

PR #2226, opened 2022-05-27
RD-4990 Add a note about `cfy nodes list --run-checks`

tehasdf:
The extended view part is true right now. Will it always be true?
Who knows.


tehasdf:
staging https://docs.cloudify.co/staging/nodes-list-run-checks-1/cli/orch_cli/nodes/

mateumann (approved):

PR #2247, opened 2022-07-14
RD-5205 part 2: Update agents page

leon-good-life:
docs update for https://github.com/cloudify-cosmo/cloudify-stage/pull/2062


qooban (commented):
Don't we need to also update description/screenshot on Agents widget page?

PR #2265, opened 2022-07-29
RD-4713 Update composer documentation

kubama:


content/developer/composer/blueprint-creation.md
```diff
@@ -62,8 +62,25 @@ Click the node to display the node property panel. It includes the following com
 
 To close the panel, press **Esc** or click the close window icon on the top-right corner.
 
-![Working with Nodes]( /images/composer/working-with-nodes.png )
+![](/images/composer/properties-pane.png "Properties pane")
```
qooban, line 65: Didn't now that it's possible to specify tooltip this way. Probably we should use it more often.

Alt name not specified intentionally?

qooban (approved):

qooban:
@kubama I suspect there's more to change in https://docs.cloudify.co/staging/dev/developer/composer/ - e.g. Inputs & Outputs was separated and Capabilities section was added. The definition of Inputs has changed and Stencils pane is a bit different (new icons). Are you planning to update it within RD-4713 as well?

content/developer/composer/blueprint-creation.md
kubama: Yes, I could use the same string for title and description, but I don't think it would make sense.

kubama:
RD-4713 is part of node properties epic, so I wasn't planning to include anything that is out of this scope.

qooban:
OK, fine. And what is the plan for documenting stuff from the other Composer epics? (for example: RD-4388, RD-4395 or RD-4821; possibly more as I'm not sure these are all we touched recently)

kubama:
I think there should be dedicated tickets for documentation update in those epics. As I can see there are none... I think I can cover these areas in this PR then.

content/developer/composer/managing-inputs-outputs.md
```diff
@@ -1,26 +1,20 @@
...
-title: Managing Inputs and Outputs
+title: Managing Inputs, Outputs and Capabilities
 category: Composer
 draft: false
 weight: 500
 aliases: /composer/managing-inputs-outputs/
 ---
  
-{{< param cfy_composer_name >}} enables you to add inputs and outputs to the Blueprint under the **Inputs & Outputs** node.
+{{< param cfy_composer_name >}} enables you to manage blueprint's [inputs]({{< relref "developer/blueprints/spec-inputs.md" >}}), [outputs]({{< relref "developer/blueprints/spec-outputs.md" >}}) and [capabilities]({{< relref "developer/blueprints/spec-capabilities.md" >}}).
```
qooban, line 10: Good idea to make it linked to spec pages.

qooban (approved):
Good idea to make screenshots not having header with version 👍🏻

PR #2266, opened 2022-08-01
RD-5503 note about cfy events list: node id and operation name

tehasdf:


tehasdf:
implemented in cloudify-cosmo/cloudify-cli#1433

geokala (approved):

PR #2287, opened 2022-08-05
Default values not for data-based types

mateumann:
Data-based types: `blueprint_id`, `deployment_id`, `secret_key`,
`capability_value`, `scaling_group`, `node_id`, `node_type` and
`node_instance` cannot have default values defined.  This is because
inputs and parameters of that types should depend on DB-content rather
than blueprint's specification.


alexmolev (approved):

mateumann:
https://docs.cloudify.co/staging/no-default-values-for-data-types-640/

PR #2290, opened 2022-08-10
Fix Python versions used in Cloudify Agents

mateumann:


mateumann:
Updated docs: https://docs.cloudify.co/staging/agent-versions/install_maintain/agents/

PR #2296, opened 2022-08-15
RD-5041 Labels Widget - not permitted keys

leon-good-life:
PR in Stage: https://github.com/cloudify-cosmo/cloudify-stage/pull/2130


leon-good-life:
In this PR I copied the error message to the documentation:
```
All labels starting with `csys-` are reserved for internal usage and should not be used unless there is a need to override the default functionality. The system will prevent overdoing labels that can sacrifice consistency. List of reserved lables: csys-location-lat, csys-env-type, csys-environment, csys-obj-name, csys-location-name, csys-location-long, csys-wrcp-services, csys-obj-parent, csys-wrcp-group-id, csys-obj-type.
```
But on second thought, I think it's unnecessary.

PR #2386, opened 2022-11-03
updated docs for upcoming widgets feature

Blewin:


Blewin:
out of date and out of scope for now

PR #2391, opened 2022-11-08
RD-6148 add arm64 links - Latest

yuvlar:
I decided to update the 'description', 'weight' according to Master - wasn't sure which of those is the most updated one.
If you prefer the original way please let me know.


yuvlar:
@alexmolev I see that there's a similar page in /cloudify_manager/releases/_index.md - should I update it as well?

PR #2394, opened 2022-11-09
RD-6206 Add Cloudify UI Components sync check script to CI pipeline

qooban:
1. Added new script to check UI components versions sync. I used `checkDocs.sh` script from `cloudify-stage` as a base and adjusted it to `docs.getcloudify.org` specificity.
2. Updated CI pipeline to run the check on `*-build` branches only.

Tested with modified Jenkinsfile in: https://jenkins.cloudify.co/blue/organizations/jenkins/Docs-Multibranch/detail/RD-6206/7/pipeline - you can see there the error thrown on versions mismatch.

Once this is merged, I'll adjust it and apply on `6.4.0-build` branch.


qooban:
> Is this check failure going to block doc PRs from merging?

Yes.

qooban:
> Is this check failure going to block doc PRs from merging?

Good point. Seems like "build&deploy-prod" does not have **Required** label.

Oh, and I realized that the check won't work as I thought it to work. 
It should actually be triggered on all branches which are intended to be merged to `*-build` branches... Otherwise we won't see it in PRs. With the current implementation we would need to go to Jenkins and see result for `*-build` branch. That's not the best I think.

Any suggestions how to make it better?

kubama:
> It should actually be triggered on all branches which are intended to be merged to *-build branches... Otherwise we won't see it in PRs. With the current implementation we would need to go to Jenkins and see result for *-build branch. That's not the best I think.

Why would we want to see this failure in PR? I can't think of any real life scenario where this failure would be caused by a change made in a PR. Thus, it should be done in `*-build` branches only. If and how this failure is going to be monitored or escalated is a different story.

alexmolev (approved):

qooban:
> Why would we want to see this failure in PR? I can't think of any real life scenario where this failure would be caused by a change made in a PR. Thus, it should be done in *-build branches only. If and how this failure is going to be monitored or escalated is a different story.

Yes and no. I'd like to see that before I merge anything to `*-build` branch, but I know it might be problematic.
Monitoring/escalation is equally important to technical solution we are introducing here IMHO, but I'm going to leave it like it is right now without digging deeper as I don't have better solution than what I did already as part of RD-6206 (at least for now).

PR #2437, opened 2022-11-28
RD-6447 new datatypes

mateumann:


mateumann:
Updated pages:
* https://docs.cloudify.co/staging/RD-6447-new-datatypes-1/developer/blueprints/spec-inputs/
* https://docs.cloudify.co/staging/RD-6447-new-datatypes-1/developer/blueprints/spec-workflows/ (just a link there)

PR #2494, opened 2023-01-18
RD-6636 Prometheus a.k.a. Status Reporter

mateumann:


mateumann:
staging documentation: https://docs.cloudify.co/staging/RD-6636-prometheus-1/cloudify_manager/architecture/monitoring/

geokala:
A few comments on the diagrams:
1. You use 'Cloudify' in there, which you should probably avoid for the same reason we have product name parameterisation. You can probably just say 'Manager'?
2. The interaction arrows on the individual component diagrams make it look at first glance like data flows in to the node_exporter and out from the <other>_exporter. It'd probably make it easier to read at first glance if the external+nginx+prometheus arrows were moved slightly inward so that they didn't line up directly with the exporter ones.
3. I think the full cluster diagram is missing the prometheus on each node.

mateumann:
ad. 1. I'm pretty sure “Cloudify” is used only in path names (e.g. `/etc/cloudify/ssl/`) or service names (e.g. `cloudify-rabbitmq`).  I see no way around it as those are implementation details not as easily swappable as e.g. {{< param product_name >}}.

ad. 2. Thanks, fixed that.

ad. 3. Did you mean the picture on the https://docs.cloudify.co/staging/RD-6636-prometheus-1/cloudify_manager/architecture/monitoring/federation/ page?  If you did, it's "Status Reporter" (of which Prometheus is a component) on all other nodes Prometheus is talking to. I put it like that because there are also nginxes (with TLS and auth) involved (and technically Prometheus is talking to nginxes).

mateumann:
ad. ad. 1 Oh, you meant the diagrams.  Let's fix that too.

alexmolev (approved):

PR #2519, opened 2023-02-21
RD-6444 Update documentation covering blueprint deployment

Vorbert-Kruk:


content/working_with/console/widgets/blueprints.md
```diff
@@ -99,11 +99,12 @@ The default visibility is "Tenant", and according to the logged-in user's permis
 
 1. Click the deploy icon ![Deploy icon]( /images/ui/icons/deploy-icon.png ).   
 2. In the Deploy Blueprint dialog, specify a name for your deployment.
-3. Specify the required deployment inputs.   
+3. Specify environment on which the blueprint should be deployed on - this option is presented only when the blueprint is using [get_environment_capability]({{< relref "developer/blueprints/spec-intrinsic-functions#get_environment_capability" >}}) intrinsic function.
```
Vorbert-Kruk, line 102: I wasn't sure if I should mark it at the beginning as `(optional)`
From one point of view this value is not always visible (so it's not required) but form the other side, it's required when presented

I'm open for suggestions regarding that aspect 😄  

If we'd agree to go in a certain direction, I'll also update other files, which are referring to this functionality 🎉

content/working_with/console/widgets/deployments.md
```diff
@@ -50,6 +50,7 @@ The hamburger menu (![list icon]( /images/ui/icons/list-icon.png ))
 on the right of every deployment allows performing the following operations:
 
 * Install/Uninstall deployment
+* Deploy blueprint on deployment - this option is only available when the deployment is containing `csys-obj-type: environment` label
```
qooban, line 53: In other places you are referring to environment, so maybe here you can do it as well:
```suggestion
* Deploy blueprint on environment - this option is only available when selected deployment is an environment
```
?

BTW I hope in the future https://cloudifysource.atlassian.net/browse/RD-4113 will be implemented, so all mentions of `environment` can be linked to proper page.

content/working_with/console/widgets/blueprints.md
qooban: I'm fine with this direction.

qooban (approved):
Left just one comment to consider.

Side note: The best would be to have someone from the product team reviewing this kind of PRs.

content/working_with/console/widgets/deployments.md
Vorbert-Kruk: I'd be fine with that direction 💪 
But the issue is that the PR has been merged before I could even consider doing it 😅 
Sorry about that

PR #2521, opened 2023-02-23
Parameterized the files under content_bestpractices

oloten:
For better integration with Wind River System's conductor, replaced or reworked product names with either variables or generic content.


oloten:
This is the parameterized content for the content/bestpractices files. I am using separate branches for each content group. These are not sequential and are each branched from the master. Please let me know if you would like me to proceed using a different branching strategy.

alexmolev (approved):

PR #2527, opened 2023-03-08
RD-6001 info about additional columns in deployments view widget

Blewin:


Blewin:
@qooban can you help with verifing that this is the only change needed?

Blewin:
because on one hand it seems small, but on the other hand, i had previously written about not being able to sort by labels, which is now false :D

qooban (approved):

PR #2532, opened 2023-03-20
RD-6607: Updated screenshots after theming changes

Vorbert-Kruk:
PR is about updating screenshots related to theming changes

As theming changes were quite extensive, I've used the strategy of going through every [static/images/ui](https://github.com/cloudify-cosmo/docs.getcloudify.org/tree/master/static/images/ui) image to update it if the update would be needed and if I'd be able to recreate presented on the screenshot circumstances

Additional benefit to that strategy is the fact that I was able to remove screenshots which weren't used anywhere within this repo 🎉


qooban (commented):
Looks very good 👍🏻 I really appreciate hard work of going over this big amount of screenshots.

I've got just one - general comment: I see you are lucky man using 4K display ;-) But looking at [stats](https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide) we probably won't benefit much from it yet and also image file sizes are bigger than before. AFAIR Hugo (at least with current configuration) is not compressing images, it just shows them as is.

What do you think about rescaling images (e.g. 50%)? There are probably some tools to do mass rescale on a list of images. Or maybe there's Hugo configuration to do that automagically? 🤔

Vorbert-Kruk:
> I've got just one - general comment: I see you are lucky man using 4K display ;-) But looking at [stats](https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide) we probably won't benefit much from it yet and also image file sizes are bigger than before. AFAIR Hugo (at least with current configuration) is not compressing images, it just shows them as is.
> 
> What do you think about rescaling images (e.g. 50%)? There are probably some tools to do mass rescale on a list of images. Or maybe there's Hugo configuration to do that automagically? 🤔


I was just cutting out some pieces of screenshots made on Dell's laptop
As those are appearing in 4k, I'm impressed 😄 
Probably the display is 1920x1024, but the resolution is 2x - that's why 4K effect may be visible, as the pixels are being automatically scalled 🤔

Anyway, I think that the proposition about rescaling images is a valid one
Because of that, I've took the opportunity to learn how to scale bunch of images in a simple way 🎉 
And the results are quite satisfying, so if in the future anybody would wonder how to perform those type of actions on windows, I'll be more than happy to help :)

Images were resized within eb65d976e9933a4ce492983a29889c0be1b9655e 🎸

qooban (approved):
Looks good! 👍🏻

